PR #1, opened 2018-12-12
Start of Docker images projects

glicht:
Base Docker images


docker/python3/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM python:3.7.1-alpine3.8
+
+COPY requirements.txt .
+
+RUN pip install --no-cache-dir -r requirements.txt
+
+RUN addgroup -g 4000 demisto \
+  && adduser -u 4000 -G demisto -D demisto -s /bin/sh
```
dorsha: it won't be always demisto demisto, it can be also socserver and rsa

docker/python3-deb/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+# Basic linux utilities
+RUN apt-get update \
+  && apt-get install -y \
+    procps
+
+COPY requirements.txt .
+
+RUN pip install --no-cache-dir -r requirements.txt
+
+RUN groupadd --gid 4000 demisto \
```
dorsha: same here

docker/dockerbuild/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM circleci/python:2.7.15-stretch
```
meirwah: why not `slim` version of stretch here too?

docker/python-deb/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+# Basic linux utilities
+RUN apt-get update \
+  && apt-get install -y \
+    procps
+
+COPY requirements.txt .
+
+RUN pip install --no-cache-dir -r requirements.txt
+
+RUN groupadd --gid 4000 demisto \
```
meirwah: did you test these docker images work well in demisto ? on various OS's ?

docker/dockerbuild/Dockerfile
glicht: This is from circleci. It is an image to use for building our images (dev only). It is not an image we will be shipping. Circle don't have a slim version from what I saw.

docker/python-deb/Dockerfile
glicht: Only tested locally (mac). Need to discuss how we want to test them. I am open for suggestions. We can take this offline.

PR #6, opened 2018-12-24
new utilities for working with word documents

idovandijk:
This is the new docker image that contains libreoffice + python-docx


idovandijk:
The docker image you helped me with looks good too! 👍

PR #12, opened 2019-01-03
add new image 'ldap' with ldap3 package

noaco:
For Active Directory BYOI


glicht:
@noaco all looks good. If you have nothing to add you can merge this into master. Then the docker image will be available at demisto/ldap

PR #13, opened 2019-01-04
Zeep

ssudheerk99:
@glicht  as requested Docker image for Zeep  python package.
 let me know for any change requests


glicht:
@ssudheerk99 thanks for submitting this. I've added a small change to use our docker image as base (demisto/python3) and changed the pip install flow. We basically first use pipenv file to generate a full requirements.txt file (our build script does this) and then use it to install. Also, we remove the dev packages that are not needed for runtime.

glicht:
The docker image is now available at: https://hub.docker.com/r/demisto/zeep/tags

Use: demisto/zeep:1.0.0.78

PR #16, opened 2019-01-14
Get pcap's http container

funerr:
Add a docker container for the `PcapHTTPExtractor` script. 

It includes the binary `tshark` and `pyshark` python lib for pcap/pcapng inspection.
Using the base image: `demisto/python3-deb:1.0.0.53`.


docker/pcap-http-extractor/Dockerfile
```diff
@@ -0,0 +1,14 @@
+
+FROM demisto/python3-deb:1.0.0.53
+
+COPY requirements.txt .
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+  gcc \
+  python-dev \
+  tshark \
```
glicht: Please install the thirdparty required debian packages in a separate RUN command as we do with: https://github.com/demisto/dockerfiles/blob/master/docker/office-utils/Dockerfile#L15

funerr: I did it in #66a1394.
BTW, why is that needed?

glicht: I added a small commit to also remove the lists. All looks good. Regarding your question it is not a really needed to separate but it makes things easier to understand with this separation, when there is need to review what the image contains.

glicht:
I approved. You can go ahead and merge if there is nothing on your side to add.

PR #17, opened 2019-01-16
Python images update

glicht:
update python base images with modified date to trigger rebuild and use latest docker base images

### DIFF
--- a/docker/dockerfiles_update_date.sh
+++ b/docker/dockerfiles_update_date.sh
@@ -1,0 +1,19 @@
+#!/usr/bin/env bash
+
+# exit on errors
+set -e
+
+#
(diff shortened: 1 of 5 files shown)


docker/dockerfiles_update_date.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+
+# exit on errors
+set -e
+
+#
+# Will go over all Dockerfiles that have a comment of the form:
```
anara123: will this script manually or automatically? Because I don't see who triggers this script

glicht: It is a utility script. To run manually when you need to update the python images as I just saw that is needed. In the future need to see how we do this automatically.

PR #39, opened 2019-03-19
Content build pyenv setup, dependencies, 2and3 image

glicht:
Properly setup pyenv and set default python versions for use from pyevn
Add pytest-mock and update dependencies
Add 2and3 image with both python2 and python3 setup

### DIFF
--- a/docker/content-build-2and3/.gitignore
+++ b/docker/content-build-2and3/.gitignore
@@ -1,0 +1,1 @@
+requirements.txt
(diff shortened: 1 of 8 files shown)


glicht:
@ronykoz want to use this image which has also dependencies installed for python3. Started using it at: https://circleci.com/gh/demisto/content/19975

ronykoz:
We want this in order to run unit-test in our build, right?

PR #41, opened 2019-03-27
Create docker image for snowflake integration

avidan-H:


Itay4:
@glicht can you have a look?

glicht:
It is failing as there is a dependency on cffi and it fails to compile it:

image

Looks like there is need to install libffi-dev. Will add this soon...

PR #43, opened 2019-04-03
update version of py3 images to 3.7.3, content build images update

glicht:
We updated the docker image to use 3.7.3 but forgot to update the version


glicht:
Will separate this into 2 prs as seems that the update for content-build is more complex.

PR #49, opened 2019-04-23
build_docker.sh: fix #48: DOCKER_ORG is reset to 'demisto' when not using CircleCI

### LINKED ISSUE #48: build_docker.sh: DOCKER_ORG is always reset to 'demisto' when not using CircleCI
Summary: Under some circumstances, build_docker.sh ignores any provided DOCKER_ORG value and always resets to 'demisto'.

**Steps to reproduce / contributing factors:**
- The script must be located in a git-tracked folder (which is true if it was downloaded using git clone)
- $CIRCLE_BRANCH must be empty

**Problem in build_docker.sh**
- Code on lines 139-141 sets CIRCLE_BRANCH to the git branch value of the local folder (i.e. master, if it was pulled directly from GitHub without further modification).
- Because the branch is then set to master, the check on line 168 succeeds, and the DOCKER_ORG is set to 'demisto' on line 182.

**Available workaround**
If the script is run with any non-null CIRCLE_BRANCH value, the provided DOCKER_ORG is honored.

**Proposed solution**
If I'm not mistaken, then $CIRCLE_SHA1 must be set in order to use the CircleCI integration. If that is indeed true, the code from lines 139-142 can just be moved into the block at lines 133-137.

flmsc:
This should fix issue #48 by only automatically setting CIRCLE_BRANCH if a CIRCLE_SHA1 value was provided at runtime.

This is built on the assumption that CIRCLE_SHA1 must be present in order for the CircleCI integration to work -- please reject if that assumption is wrong.

Cheers!


glicht:
Thanks for submitting this. We will review and get back soon.

glicht:
thanks for submitting this. I've put in a small change to simplify things.

PR #61, opened 2019-05-13
Fix for rebuild + pyenv optional change

glicht:
Fixed so when doing a rebuild on master we don't build all images. 

Clarified that pyenv is optional as long as you have python 2 +3 installed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,9 +12,11 @@ The script `docker/build_docker.sh` is used to build all modified docker images.
 
 ## Getting Started
 **Pre-requisites:**
-* Install pyenv:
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -12,9 +12,11 @@ The script `docker/build_docker.sh` is used to build all modified docker images.
 
 ## Getting Started
 **Pre-requisites:**
```
Itay4: shouldn't homebrew be the first prerequisite ?

glicht: I think it's ok. Don't want to make this mac specific too much.

PR #66, opened 2019-05-27
Create Encoding Detection docker image

amshamah419:
Needed for ReadFile V2


glicht:
I don't think we need this. chardet is already included with base images.

amshamah419:
Agreed. I'll close.

PR #67, opened 2019-05-28
Add sane doc reports

agamm:
Used for the automation that produces docx reports.

**Currently should be in dev mode**.


glicht:
@agamm build is failing. It seems that sane_src is empty.

meirwah:
@agamm plz note build fails here

Pipfile.lock
```diff
@@ -0,0 +1,20 @@
+{
```
glicht: Unnecessary file. Please delete.

Pipfile
```diff
@@ -0,0 +1,11 @@
+[[source]]
```
glicht: Unnecessary file. Please delete.

PR #68, opened 2019-05-28
Slack docker

liorblob:


meirwah:
@liorblob what is this pending on ?

liorblob:
@meirwah We were not sure if the python client is fit for the integration. I'll merge now

PR #73, opened 2019-06-07
cloaken docker update; add response_status to return data

ajcypherint:


glicht:
Nothing is being updated except the version. Don't really see why it is needed.

ajcypherint:
I need the docker image rebuilt. I made a change to cloakensdk. Is there
another way to do this? Or do these all rebuild nightly?

glicht:
You need to run: `pipenv update` in the `cloaken` directory. 

This will update the Pipfile.lock file with the latest package versions. Then commit this file.

ajcypherint:
Thank you!

glicht:
Are you pushing another commit of Pipfile.lock?

ajcypherint:
Yes will do tonight.

ajcypherint:
Ok I updated the Pipfile.lock with the latest version of CloakenSDK

PR #78, opened 2019-06-19
Create Boto3 for AWS on Python3

amshamah419:
Enabling forward compatibility with creation of Python3 compatible boto3 image.


glicht:
@amshamah419 I've pushed an exclude to docutils regarding the license failure.

PR #277, opened 2019-07-29
Dockerfile for uptycs integration

bkschmoll-uptycs:


glicht:
@bkschmoll-uptycs thanks for the contribution. I've pushed a commit with a few small changes. Once the build completes and passes, I will merge.

bkschmoll-uptycs:
Thanks!

PR #281, opened 2019-07-31
nltk docker

glicht:
docker image for nltk

Relates to: https://github.com/demisto/etc/issues/17097


glicht:
@erezh31 please take a look. this image is for nltk update with install supporting non-root user.

PR #329, opened 2019-08-06
requests_oauthlib image

orenzohar:


docker/oauthlib/Dockerfile
```diff
@@ -0,0 +1,11 @@
+
+FROM demisto/python3:3.7.4.1150
```
glicht:
```suggestion
FROM demisto/crypto:1.0.0.1024
```

If you use our crypto docker as a base you don't need to install libffi and openssl and the dockers shipped are smaller and optimised.

docker/oauthlib/Dockerfile
```diff
@@ -0,0 +1,11 @@
...
+
+COPY requirements.txt .
+
+RUN apk --update add --no-cache libffi openssl
```
glicht: after adding: `FROM demisto/crypto:1.0.0.1024` you can remove this.

PR #333, opened 2019-08-07
rasterize chromium image

orenzohar:


docker/rasterize-chrome/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+&& pip install --no-cache-dir -r requirements.txt \
+&& apt-get purge -y --auto-remove \
+  gcc \
+  python-dev \
+&& rm -rf /var/lib/apt/lists/*
+
+# Install Chrome
+RUN apt-get update
+RUN wget https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb
+RUN dpkg -i google-chrome-stable_current_amd64.deb; apt-get -fy install
```
glicht: Combine the above three commands into one and add cleanup. Also, not sure about the apt-get install part and if it is needed.

orenzohar:
@glicht seems like `wget` fails in getting the latest chrome from google (locally too, looks like a ban for too many requests). I will upload the zip into the docker directory for now

docker/rasterize-chrome/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+  python-dev \
+  wget \
+  libxss1 \
+  libappindicator1 \
+  libindicator7 \
+&& pip3 install --no-cache-dir -r requirements.txt
+
+# Install Chrome
+#RUN wget https://dl.google.co/linux/direct/google-chrome-stable_current_amd64.deb; \
+RUN dpkg -i google-chrome*.deb; \
```
glicht: Use `&&` and not `;`

orenzohar: && will not run the next command on an exit code other than 0

docker/chromium/Dockerfile
```diff
@@ -1,8 +1,26 @@
...
+  python-dev \
+  wget \
+  libxss1 \
+  libappindicator1 \
+  libindicator7 \
+&& pip3 install --no-cache-dir -r requirements.txt
+
+# Install Chrome
+#RUN wget https://dl.google.co/linux/direct/google-chrome-stable_current_amd64.deb; \
+RUN dpkg -i google-chrome*.deb; \
```
glicht: Use `&&` instead of `;`

.dependabot/config.yml
```diff
@@ -502,3 +502,14 @@ update_configs:
         dependency_name: demisto/*
         update_type: all
 
+  - package_manager: python
```
glicht: Not needed as you've changed the name of the project to am existing one.

docker/chromium/Dockerfile
orenzohar: > && will not run the next command on an exit code other than 0 (which is expected in this scenario)
@glicht

glicht: ok. so use && for the next steps. and also add a verification that the install actually installed chrome. You can run something of the form: `/opt/google/chrome/chrome --version | grep 76.0.3809`

docker/chromium/Dockerfile
```diff
@@ -1,8 +1,26 @@
...
+apt --fix-broken -y install; \
+apt-get -y install unzip
 
-RUN apk --update add --no-cache --virtual .build-dependencies python-dev build-base wget git \
-  && pip install --no-cache-dir -r requirements.txt \
-  && apk del .build-dependencies
+# Clean Up
```
glicht: Don't do the cleanup in a separate RUN. It should be done within each separate RUN.

docker/chromium/Dockerfile
orenzohar: > `&& python3 -c "from webdriver_manager.chrome import ChromeDriverManager; ChromeDriverManager().install()" \`

I added a python command that will ensure that chrome installed successfully, and will init the web driver so the

orenzohar:
@glicht updated the PR for the review

docker/chromium/Dockerfile
orenzohar: if I do the clean up before the dpkg, the installation will fail

docker/chromium/Dockerfile
glicht: Excellent.

docker/chromium/Dockerfile
glicht: I added a commit regarding what I meant about cleanup. You can see that now the image on docker hub is smaller.

PR #338, opened 2019-08-08
Add docker image for Google vision API

pingunaut:
PR for initial support of Google Vision API is in preparation. This Image is the foundation for it.


glicht:
Hi @pingunaut thanks for submitting this. I see you are using the Debian based image. Did you try with an Alpine based image? We usually prefer Alpine as it is a smaller base image.

pingunaut:
Hi @glicht thanks a lot for your quick response. Yes I tried that one before but gave up after some hours. The python dependencies are missing some libraries on alpine like libstdc++.so.6

I tried to add libstdc++ libc6-compat and a view other packages but it did not work out well :(

Using the debian based image, it just works out of the box

glicht:
Ok. Sounds good.

PR #412, opened 2019-08-21
store image artifacts

glicht:
When building an external pr (from a fork) store the resulting docker image as an artifact in circle.


glicht:
Artifacts will be stored only for external builds. Example build before this enforcement:

https://circleci.com/gh/demisto/dockerfiles/1699#artifacts/containers/0

PR #420, opened 2019-08-27
Google Cloud Storage image

panro:
Docker image for Google Cloud Storage integration
Content PR: https://github.com/demisto/content/pull/4146


.dependabot/config.yml
```diff
@@ -523,3 +523,14 @@ update_configs:
     - match:
         dependency_name: demisto/*
         update_type: all
+  - package_manager: python
```
Itay4: i dont think you need to add it manually
@glicht can you confirm ?

glicht: It is ok. The create docker image script adds this. We need this for dependabot to detect updates.

PR #440, opened 2019-09-02
update py package and add ansible to content build image

glicht:


glicht:
Update of py3 packages will be done after first build of base image

yaakovi:
can we trigger a content build with the new docker before merging it?

PR #442, opened 2019-09-03
Added fips compiler for Demisto

BenJoParadise:
Used to compile Demisto in a fips-certified manner.


BenJoParadise:
@glicht i think something needs to be added to `.dependabot`, but not sure what

glicht:
To add the config run: ./docker/add_dependabot.sh docker/fips-compiler

We print this as part of the output of the failed build.

BenJoParadise:
Not needed, we are taking a different approach.

PR #451, opened 2019-09-09
use demisto-py 2.0 from pypi

glicht:


glicht:
@amshamah419 are we good?

amshamah419:
I think the docker image is ready. I'm still debugging the build process however. So long as the content build is using the old image until we are ready to merge the new build process, we should be okay.

PR #458, opened 2019-09-11
nmap 

Itay4:


glicht:
I am pretty sure you need to install nmap. Add the following:
```
RUN apk --update add --no-cache nmap
```

PR #587, opened 2019-10-02
Chromium pdf to image libs

orenzohar:
added required libraries for image rendering from pdf files.


docker/chromium/Pipfile
```diff
@@ -8,6 +8,9 @@ verify_ssl = true
 [packages]
 selenium = "*"
 webdriver-manager = "*"
+PyPDF2 = "*"
+pdf2image = "*"
+numpy= "*"
```
Itay4:
```suggestion
numpy = "*"
```

PR #588, opened 2019-10-02
chromium: change to download the zip and deb files to avoid copy

glicht:
Trying to reduce image size by avoiding the COPY step


glicht:
Reduces size from 357 to 295:

image

PR #589, opened 2019-10-02
bring back no-install-recommends to office-utils

glicht:


glicht:
size reduced to 272 MB

PR #604, opened 2019-10-04
Update sane-doc-reports - fix investigation reports

agamm:
These are fixes for the investigation reports, so merging ASAP would be awesome (before Sunday).

Related to issues:
https://github.com/demisto/etc/issues/19429
https://github.com/demisto/etc/issues/19306
https://github.com/demisto/etc/issues/19298


agamm:
Thanks @glicht :)

PR #625, opened 2019-10-15
Change docker image base

GalRabinDemisto:
@yaakovi 
Hi,
Open a new PR after our conversation


docker/content-build-2and3/Pipfile
```diff
@@ -21,6 +21,9 @@ ansible-runner = "*"
 rsa = "*"
 paramiko = "*"
 boto3 = "*"
+pyminizip = "*"
+beautifulsoup4 = "*"
+pypdf2 = "*"
```
yaakovi: is that case-insensitive?

GalRabinDemisto: @yaakovi 
Yes its case insensitive, It created by the:
`pipenv install PyPdf2`
[Check it out](https://stackoverflow.com/questions/26503509/is-pypi-case-sensitive)

PR #787, opened 2019-11-02
new content build without dependencies and pyenv

glicht:
Trying a new approach for the content build where we use virtualenv and circle ci caching. Thus, reducing the build of our build image.


yaakovi:
what about the other docker image (content-build-2and3)? why we are keeping it?

glicht:
Keeping it for history. In case we want to go back.

PR #881, opened 2019-11-14
add gmail docker image

adi88d:
Pyhon 3.7 version docker image with the packages:
google-api-python-client and cryptography.


glicht:
Let's name this image: `google-api-py3`. I don't see here anything gmail specific. Note: that we have an image of `google-api` but it is py2. 

Also, this image should be based off `demisto/crypto` as we do with teams image.

docker/gmail/Dockerfile
```diff
@@ -0,0 +1,13 @@
+
+# Note: use this image as a base if you are dependent upon cryptography
```
glicht: Remove these comments and base off `demisto/crypto`.

docker/google-api-py3/Dockerfile
```diff
@@ -0,0 +1,13 @@
+
+# Note: use this image as a base if you are dependent upon cryptography
+# Make sure to include in your Pipfile: cryptography, cffi. So these dependencies
+# will be managed by dependabot and upto date.
+# See teams image for an example.
```
glicht: You can remove the above comments

PR #1118, opened 2019-12-12
add akamai docker

GalRabin:


glicht:
Use `demisto\crypto` as a base. Since you are dependent upon cryptography.

docker/akamai/Pipfile
```diff
@@ -0,0 +1,18 @@
...
+name = "pypi"
+
+[packages]
+edgegrid-python = "*"
+requests = "*"
+olefile = "*"
+python-dateutil = "*"
+dateparser = "*"
+pyyaml = "*"
+funcy = "*"
```
glicht: No need for all the dependencies that are already included in the base image. You should add only `edgegrid-python`

GalRabin: forgot to change it when I change the base
Thanks

PR #1138, opened 2019-12-18
matplotlib docker image

Itay4:


Itay4:
@glicht see my commit, had to add libstdc++ since got the following error:
```
ImportError: Error loading shared library libstdc++.so.6: No such file or directory (needed by /usr/local/lib/python3.7/site-packages/matplotlib/ft2font.cpython-37m-x86_64-linux-gnu.so)
```

PR #1159, opened 2019-12-25
Updated taxii image to python 3

DeanArbel:


glicht:
@DeanArbel let's use as a base the following image: demisto/btfl-soup and then you also don't need to specify this dependency in the pipfile.

Add a verify.py script just to check the lxml stuff. I've seen problems with this dependency.

docker/taxii/.gitignore
```diff
@@ -1 +0,0 @@
-requirements.txt
```
glicht: You should leave this file to avoid checking in by mistake requirements.txt

glicht:
@DeanArbel see my commit. I fixed the runtime deps and removed the build deps when doing installing via pip. Image size is now down to 56MB (from 134MB).

PR #1164, opened 2020-01-02
Add fast text docker image

erezh31:


docker/fasttextv2/Dockerfile
```diff
@@ -0,0 +1,19 @@
+
+FROM demisto/python-deb:2.7.17.4583
+
+COPY requirements.txt .
+
+RUN apt-get update && apt-get install -y zlib1g-dev libicu-dev g++
+
+RUN pip install --upgrade pip
+
+RUN apt-get update && apt-get install -y gcc g++
```
glicht: Do you need g++ and gcc during runtime? If only needed during build then the below command will install it before doing the `pip install...`. and remove after.

docker/fasttextv2/Dockerfile
```diff
@@ -0,0 +1,19 @@
+
+FROM demisto/python-deb:2.7.17.4583
+
+COPY requirements.txt .
+
+RUN apt-get update && apt-get install -y zlib1g-dev libicu-dev g++
```
glicht: Remove this. Add dev libs to the step where we install python-dev and later remove it.

glicht:
@erezh31  see my changes. I am merging the image and it will be available soon.

PR #1282, opened 2020-01-28
add kerberos to winrm

glicht:
## Status
Ready

## Description
add kerb5 to winrm


glicht:
@aburt-demisto please use the image: demisto/winrm:1.0.0.5794

PR #1365, opened 2020-02-24
powershell deb base image + version bump

glicht:
## Status
Ready

## Related Issues
related: https://github.com/demisto/etc/issues/22231

## Description
Debian based base image for powershell


yardensade:
Don't we need to pre-install openssh-client package in the docker file?

glicht:
That will be in another docker. This is just adding a new base docker based upon debian.

PR #1456, opened 2020-03-12
Sane-pdf-reports

agamm:
## Status
Ready

## Related Issues
related: https://github.com/demisto/etc/issues/22495

## Description
Moving the pdf section of sane-reports to be docker'ized.


agamm:
Creating a new branch so old files won't be included in history.

PR #1491, opened 2020-03-16
Add a python helper to download circleci artifacts

agamm:
## Status
Ready 

## Related Issues
(related to future PR on sane-pdf-reports where we will need to fetch the latest artifacts from circleci).

## Description
This is a python helper utility script everyone can use for downloading CircleCI artifacts.


agamm:
We will move to the specific directory.

PR #1499, opened 2020-03-18
add readme validate dependencies

GalRabin:
## Status
In Progress

## Description
Adding the following for docker build image:
1. `package.json` with dependencies for validating readme.
2. setting the variable to true.

### DIFF
--- a/docker/content-build/package.json
+++ b/docker/content-build/package.json
@@ -1,0 +1,15 @@
+{
+  "name": "content-build",
+  "version": "1.0.0",
+  "description": "Readme validation dependecies",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "keywords": [],
+  "dependencies": {
+    "@mdx-js/mdx": "^1.5.7",
+    "commander": "^5.0.0",
+    "fs-extra": "^8.1.0",
+    "jsdoc-to-markdown": "^5.0.2"
+  }
+}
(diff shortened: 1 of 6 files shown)


docker/content-build/package.json
```diff
@@ -0,0 +1,15 @@
...
+  "description": "Readme validation dependecies",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "keywords": [],
+  "dependencies": {
+    "@mdx-js/mdx": "^1.5.7",
+    "commander": "^5.0.0",
+    "fs-extra": "^8.1.0",
+    "jsdoc-to-markdown": "^5.0.2"
```
bakatzir: why is this needed?

bakatzir:
@orhovy FYI

docker/content-build/package.json
bakatzir: was beforehand.

bakatzir:
We need to create a separated docker for the demsito-sdk.

glicht:
@bakatzir @GalRabin I think it will be better to manage this in .circleci/config.yml as we do for the python deps. We reached the conclusion that doing it in docker introduces too much overhead.

bakatzir:
fixes: https://github.com/demisto/etc/issues/21123

GalRabin:
> @bakatzir @GalRabin I think it will be better to manage this in .circleci/config.yml as we do for the python deps. We reached the conclusion that doing it in docker introduces too much overhead.

https://github.com/demisto/demisto-sdk/pull/262 - @glicht  Done as you recommended

PR #1503, opened 2020-03-18
New Docker for the TAXII2 Client in Python

aburt-demisto:
## Status
Ready

## Related Issues
N/A

## Description
Docker container based on Alpine that provides the Pythong libarary STIX2 and TAXII2-client.


aburt-demisto:
Tested devdemisto/taxii2:1.0.0.6790 with my MITRE Integration using taxii2 client and all works fine so far.

glicht:
@aburt-demisto good job. Image went down to 41MB compared to 67MB of debian.

PR #1663, opened 2020-04-21
Grids -numpy and pandas

GalRabin:
## Status
Ready

## Description
In order to create the grids inefficient way, there is already an image that includes NumPy but without pandas, which usually goes together.


GalRabin:
> dependabot config has the docker/grid twice

@glicht Fixed

PR #1694, opened 2020-04-22
Threatconnect image

liorblob:
## Status
Ready

## Description
threatconnect-sdk doesn't have a proper docker image and no tags.


glicht:
What about this pr: https://github.com/demisto/dockerfiles/pull/1661 ?

liorblob:
@glicht it's python 3, this PR is python 2 (for the old integration).

PR #1723, opened 2020-04-25
Updated sdk version and docker image version

edwinabot:
## Status
Ready

## Related Issues
This is preparation for Tuesday 28th release

## Description
A series of updates and bugfixes where made to the TruSTAR SDK which are necessary for Tuesday 28th release


glicht:
@edwinabot thanks for submitting this. 

A few comments:

* The install of trustar-python is being done from master. I would like to suggest using a tag on your [trustar-python](https://github.com/trustar/trustar-python) repo and pointing to the tag in the Pipenv file used here. This will make sure that when we update the docker image for non-related dependencies trustar library is not updated. Otherwise every time we will update the image we will use a new version of trustar-python form master. We have an auto process of updating images to keep them with them updated with the latest os and python dependencies.
* The version of the image should remain with a numbered version and not a date based one. If you want to increment the version you can for example use: `2.0.0`. It better matches the way we version other images and as  we auto update the image for os dependencies and other python package dependencies the date based version losses context and confuses.

docker/trustar/build.conf
```diff
@@ -1 +1 @@
-version=1.0.0
+version=20.0.0
```
glicht, line 1:
```suggestion
version=2.0.0
```

glicht:
@edwinabot ok. We will stay with master. Regarding version: I think you have a typo. See my comment.

docker/trustar/build.conf
edwinabot: This is not a typo. See: https://hub.docker.com/r/demisto/trustar/tags

Theres a docker image published by itay44 a year ago, version number 18.12.1. This image is currently tagged as `latest`, this is wrong I need a version number higer that 18.12.1 so the docker image we are building in this PR becomes the latest.

image

glicht: Ok. We will stay with 20.0.0.

PR #1762, opened 2020-05-06
Pykafka - changed docker to use Pykafka's master branch and not release

guykeller:
## Status
Ready

## Description
This change will fix a bug regarding lz4 coding, which was fixed in Pykafka's master but never released.

### DIFF
--- a/docker/pykafka/Dockerfile
+++ b/docker/pykafka/Dockerfile
@@ -3,8 +3,8 @@ FROM demisto/python:2.7.17.5426
 
 COPY requirements.txt .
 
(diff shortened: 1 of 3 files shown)


ronykoz:
Let's add git to the docker file to make the build pass

glicht:
We should pin to a specific commit. I will to a small update to the PR.

glicht:
@guykeller don't add entries directly to the Pipfile file. You should run in the pykafak directory:
```
pipenv install xxhash
```
That way the Pipflle.lock file is updated too.

guykeller:
Closing since this is obsolete.

PR #1794, opened 2020-05-14
TruSTAR SDK version update

edwinabot:
## Status
Ready

## Related Issues
An update on TruSTAR SDK requires this changes 

## Description
Some attributes have changed on our phishing triage endpoints so this requires the version bump

### DIFF
--- a/docker/trustar/Pipfile
+++ b/docker/trustar/Pipfile
@@ -6,7 +6,7 @@ verify_ssl = true
 [dev-packages]
 
 [packages]
-trustar = {git = "https://github.com/trustar/trustar-python.git", editable = true, ref = "master"}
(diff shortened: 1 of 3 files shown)


docker/trustar/Pipfile
```diff
@@ -6,7 +6,7 @@ verify_ssl = true
 [dev-packages]
 
 [packages]
-trustar = {git = "https://github.com/trustar/trustar-python.git", editable = true, ref = "master"}
+trustar = {git = "https://github.com/trustar/trustar-python.git", editable = true, ref = "87154b98dc36c8c5a3aa9a6807be5ebdaf38ff13"}
```
edwinabot, line 9: This is not a secret but a commit hash. We cannot wait to the release to happen in PyPi.

PR #1834, opened 2020-05-21
py3ews - changed exchangelib to 3.2.0

DeanArbel:
## Status
Ready

## Description
Exchangelib 3.2.0 is required for running EWS O365 integration.

### DIFF
--- a/docker/py3ews/Pipfile
+++ b/docker/py3ews/Pipfile
@@ -6,7 +6,7 @@ verify_ssl = true
 [dev-packages]
 
 [packages]
-exchangelib = "*"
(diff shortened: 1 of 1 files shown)


glicht:
Not sure I understand why you added this. The image already has exchangelib 3.2.0.  You can run:
```bash
docker run --rm -it demisto/py3ews:1.0.0.7974 pip freeze | grep exchange
```
I get:
```
exchangelib==3.2.0
```

DeanArbel:
@glicht you're right - I was using an older version of the docker which was using 3.1.1

PR #2010, opened 2020-06-08
Update py42

unparalleled-js:
## Status
Ready

## Related Issues
n/a

## Description
The creation of an updated py42 docker container. Version pinning will soon no longer necessary (will change dependable in a separate PR)


unparalleled-js:
This is the plan, as far as I know.
1. Once this merges, there will be a new docker container available for pulling.
2. Open PR that has Code42 App upgrades / updates and change the docker container to pull new one.
3. Open another PR on this repo for remove dependable lines 1027-1028 that pin the version of py42 (so we will no longer need pinning).

Are there any whole in this plan? Does this make sense?

docker/py42/Pipfile
```diff
@@ -12,7 +12,7 @@ python-dateutil = "*"
 dateparser = "*"
 pyyaml = "*"
 funcy = "*"
-py42="==0.4.2"
+py42 = "*"
```
unparalleled-js, line 15: Does this affect the current docker image tag? I don't think it does but want to be sure. 

The current Code42 integration required a pinned version of 0.4.2.  There is a place in the dependabot yml file that has a comment near it indicating it is responsible for keeping the version pinned. After this PR merges, I will update the integration to use the latest py42 docker image and then open another PR here that removes the pinning from the dependabot yml file.

glicht: It doesn't affect the tag. The tag is controlled in the build.conf file: https://github.com/demisto/dockerfiles/blob/master/docker/py42/build.conf . Which I see you modified. So it looks like things are good.

glicht:
You can remove the lines 1027-1028 in .dependabot/config.yml already in this PR and then I will merge.

docker/py42/Pipfile
unparalleled-js: OK, so by modifying `build.conf`, it creates a new tag, and the current app goes on using the old one until we update it to use this new one. That is my understanding anyway!

unparalleled-js:
> You can remove the lines 1027-1028 in .dependabot/config.yml already in this PR and then I will merge.

As long as the nightly build does not update py42 in the current container with tag prefix 0.4.2, then we are good. Thanks!

docker/py42/Pipfile
glicht: Correct

unparalleled-js:
> You can remove the lines 1027-1028 in .dependabot/config.yml already in this PR and then I will merge.

**https://github.com/demisto/content/compare/master...code42:bugfix/INTEG-1093-handle-outside-trusted-domain-exfiltration?expand=1#diff-90cbbd278f99027bcc4102fd6e96ad4bR287**

^ This is what I didn't want to happen.

PR #2078, opened 2020-06-16
Updating sane-pdf-reports

agamm:
## Status
Read

## Related Issues
related: https://github.com/demisto/etc/issues/23059

## Description
Updates the sane-pdf-reports

### DIFF
--- a/docker/sane-pdf-reports/get_build_artifacts.py
+++ b/docker/sane-pdf-reports/get_build_artifacts.py
@@ -17,6 +17,9 @@ Usage example
 
 Use like so:
 python get_build_artifacts.py --token <token> --project <project name> --branch master --filter "regex"
+
+in sane-pdf-reports just run the docker file again.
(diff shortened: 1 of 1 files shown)


docker/sane-pdf-reports/get_build_artifacts.py
```diff
@@ -35,11 +37,12 @@ def get_args():
     return vars(parser.parse_args())
 
 
-def send_request(url, params):
-    headers = {'Accept': 'application/json'}
+def send_request(url):
+    headers = {
+        'Accept': 'application/json',
+        'Circle-Token': os.getenv(ENV_TOKEN_KEY, "")
```
Bars92, line 43: do we need token here? sane reports is public as well as circle artifacts
image

agamm: Nope :)
But if we will need it will work.

PR #2240, opened 2020-07-22
bump py3 base docker images to 3.8.5

Itay4:
## Status
Ready

## Description
bumping py3 to 3.8.5 which includes a number of security-related fixes
https://docs.python.org/release/3.8.5/whatsnew/changelog.html#python-3-8-5-final

### DIFF
--- a/docker/python3/Dockerfile
+++ b/docker/python3/Dockerfile
@@ -1,5 +1,5 @@
-# Last modified: Thu, 28 May 2020 22:53:09 +0000
-FROM python:3.8.3-alpine3.11
+# Last modified: Wed, 22 Jul 2020 08:27:22 +0000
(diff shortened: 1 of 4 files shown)


docker/python3/Dockerfile
```diff
@@ -1,5 +1,5 @@
-# Last modified: Thu, 28 May 2020 22:53:09 +0000
-FROM python:3.8.3-alpine3.11
+# Last modified: Wed, 22 Jul 2020 08:27:22 +0000
+FROM python:3.8.5-alpine3.11
```
glicht, line 2: Let's also bump the alpine version to 3.12
```suggestion
FROM python:3.8.5-alpine3.12
```

PR #2301, opened 2020-08-05
commit sklearn docker -  remove .tar

yguezpa:
## Status
Ready

## Related Issues
related: [Research] MTH Alert Prioritization #26297

## Description
This docker is basically built for sklearn package. It will be use to run automation for MTH incident prioritization model. 
It had only python with some basic ML packages: sklearn, pandas, numpy


docker/sklearn/Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM demisto/python3:3.7.5.5420
```
glicht, line 1: Why not use our latest image based on python 3.8.3?

docker/sklearn/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+		make  rsync \
+        libpng-dev freetype-dev libexecinfo-dev libgomp lapack-dev \
+		libgcc libquadmath musl  \
+		libgfortran \
+        gcc \
+        python py-pip  \
+        gfortran \
+        musl-dev \
+  && apk add --no-cache libstdc++  \
+  && apk add --no-cache openblas-dev  \
```
glicht, line 16: I think you can move this up to be part of the install which is for `.build-dependencies`

docker/sklearn/Dockerfile
yguezpa: The model is trained using python3.7. Using python 3.8 can cause come mistakes. @glicht

docker/sklearn/Dockerfile
yguezpa: I can't. If I remove those 2 packages the code won't work. 
libstdc++ is necessary for pandas 
openblas-dev for skikit learn 
@glicht

glicht: I am not too familiar with this. But `openblas-dev` is a dev pacakge. For runtime you should probably be using `openblas`. I would try adding `openblas-dev` to the build-deependencies and installing separately `openblas`.

yguezpa:
@glicht i removed the package openblas-dev
The size is now 165MB

docker/sklearn/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+COPY requirements.txt .
+
+RUN apk --update add --no-cache --virtual .build-dependencies python-dev build-base wget git g++ python3-dev\
+        openssh curl ca-certificates openssl less htop \
+		make  rsync \
+        libpng-dev freetype-dev libexecinfo-dev libgomp lapack-dev \
+		libgcc libquadmath musl  \
+		libgfortran \
+        gcc \
+        python py-pip  \
```
glicht, line 12: I don't think this is needed

docker/sklearn/Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM demisto/python3:3.8.3.9324
+
+COPY requirements.txt .
+
+RUN apk --update add --no-cache --virtual .build-dependencies python-dev build-base wget git g++ python3-dev\
+        openssh curl ca-certificates openssl less htop \
+		make  rsync \
+        libpng-dev freetype-dev libexecinfo-dev libgomp lapack-dev \
```
glicht, line 8: libgomp you are installing later on. So not needed here

yguezpa:
@glicht i added the verify.py script. Should it contains something else ?

glicht:
> @glicht i added the verify.py script. Should it contains something else ?

@yguezpa I would add: `sklearn.show_versions()`

PR #2378, opened 2020-08-11
add dill package

yguezpa:
## Status
Ready/In Progress/In Hold Ready

## Related Issues
related: 26297

## Description
Added a new package to the existing docker sklearn


docker/sklearn/Pipfile
```diff
@@ -10,6 +10,8 @@ numpy = "*"
 scikit-learn = "*"
 pandas = "*"
 nltk = "*"
+dill = "*"
+eli5 = "==0.10.1"
```
glicht, line 14: Why specific version? Why not *?

yguezpa:
> You added Pipfile and Pipfile.lock in the root dir that need to be removed

@glicht I need to remove Pipfile and Pipfile.lock and after install eli5 package?

Pipfile
```diff
@@ -0,0 +1,13 @@
+[[source]]
```
glicht, line 1: This file seems to have been checked in by mistake

Pipfile.lock
```diff
@@ -0,0 +1,196 @@
+{
```
glicht, line 1: this file seems to have been checked in by mistake

Pipfile
```diff
@@ -6,10 +6,8 @@ verify_ssl = true
 [dev-packages]
 
 [packages]
-numpy = "*"
-scikit-learn = "*"
-pandas = "*"
-nltk = "*"
+eli5 = "==0.10.1"
+dill = "==0.3.2"
```
glicht, line 10: Why do you have an exact version and not star? I would recommend changing to * so the package will be auto updated by depandabot.

glicht:
Was handled in a different pr

PR #2404, opened 2020-08-14
Trustar dockerfile update

hernantain:
## Status
Ready

## Related Issues
related: link to the issue

## Description
Bounding docker image to trustar's SDK master branch to update on every release

PR #2413, opened 2020-08-16
Handle unknown license

Itay4:
## Status
Ready

## Description
as seen in https://app.circleci.com/pipelines/github/demisto/dockerfiles/4844/workflows/e7671d1c-d02f-416e-955d-c54bdeb61d07/jobs/10720/parallel-runs/0/steps/0-106, `pip show` returned `UNKNOWN` license

added handling of that case by querying the package github repo to get the license from there

related PR: https://github.com/demisto/dockerfiles/pull/2404

### DIFF
--- a/docker/verify_licenses.py
+++ b/docker/verify_licenses.py
@@ -123,10 +123,25 @@ def check_python_license(docker_image: str, licenses: dict, ignore_packages: dic
                               docker_image, "pip", "show", name]
             pip_show = subprocess.check_output(
                 docker_cmd_arr, universal_newlines=True)
+            homepage = ''
             for line in pip_show.splitlines():
+                if line.startswith("Home-page:"):
+                    homepage = line.split(' ')[1].strip()
                 if line.startswith("License:"):
-                    print("{}: found license from pip show: {}".format(name, line))
-                    found_licenses.append(line)
+                    if 'UNKNOWN' in line:
+                        print("Got UNKNOWN license from pip show, trying to query package GitHub homepage {} to get license details.".format(homepage))
+                        owner_and_repo = homepage.split('https://github.com/')[1]
+                        repo_license = req_session.get(
+                            "https://api.github.com/repos/{}/license".format(owner_and_repo),
+                            headers={"Accept": "application/vnd.github.v3+json"},
+                            verify=False
+                        ).json()
+                        license_name = repo_license.get('license', {}).get('name')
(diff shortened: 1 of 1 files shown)


docker/verify_licenses.py
```diff
@@ -123,10 +123,25 @@ def check_python_license(docker_image: str, licenses: dict, ignore_packages: dic
...
-                    found_licenses.append(line)
+                    if 'UNKNOWN' in line:
+                        print("Got UNKNOWN license from pip show, trying to query package GitHub homepage {} to get license details.".format(homepage))
+                        owner_and_repo = homepage.split('https://github.com/')[1]
+                        repo_license = req_session.get(
+                            "https://api.github.com/repos/{}/license".format(owner_and_repo),
+                            headers={"Accept": "application/vnd.github.v3+json"},
+                            verify=False
+                        ).json()
+                        license_name = repo_license.get('license', {}).get('name')
```
anara123, line 139: what if repo don't contain a license?

docker/verify_licenses.py
```diff
@@ -123,10 +123,25 @@ def check_python_license(docker_image: str, licenses: dict, ignore_packages: dic
...
                 if line.startswith("License:"):
-                    print("{}: found license from pip show: {}".format(name, line))
-                    found_licenses.append(line)
+                    if 'UNKNOWN' in line:
+                        print("Got UNKNOWN license from pip show, trying to query package GitHub homepage {} to get license details.".format(homepage))
+                        owner_and_repo = homepage.split('https://github.com/')[1]
+                        repo_license = req_session.get(
+                            "https://api.github.com/repos/{}/license".format(owner_and_repo),
+                            headers={"Accept": "application/vnd.github.v3+json"},
+                            verify=False
```
glicht, line 137: We need to be secure by default.
```suggestion
                            verify=True
```

docker/verify_licenses.py
glicht: Better to add a default of "NOT_FOUND_IN_GITHUB"
```suggestion
                        license_name = repo_license.get('license', {}).get('name', 'NOT_FOUND_IN_GITHUB')
```

Itay4: basically `None` will be added to the the `found_licenses` and later on will count as non approved
but to make it more clear, i added a condition to check if we found a license and only then add it

PR #2414, opened 2020-08-16
update crypto based images

glicht:
## Status
Ready

## Related Issues
related: link to the issue

## Description


docker/oauthlib/Pipfile
```diff
@@ -9,8 +9,6 @@ verify_ssl = true
 requests-oauthlib = "*"
 jwt = "*"
 pyjwt = "*"
-cryptography = "*"
```
anara123, line 12: why these removed from here? because python3 contains them as core modules?

anara123 (approved):

docker/oauthlib/Pipfile
glicht: It is included in the base crypto image

PR #2590, opened 2020-09-21
adding docker image for dl-languages

lperryp:
## Status

## Related Issues
related: link to the issue

## Description
Creating a separate docker image for dl which supports languages


lperryp:
@glicht - the validation fails because we use our own library which the script can't find a licence for it

glicht:
@lperryp is this still relevant following the discussion we had?

lperryp:
> @lperryp is this still relevant following the discussion we had?

@glicht yes, but will update base image to new docker that we built

lperryp:
@glicht the I get an exception 
`Exception: es-core-news-sm: no approved license found for license: License: GPL` 
It's a download of a model from a package called `spacy` that we use in other dockers, can we overcome this?

glicht (approved):

PR #2611, opened 2020-09-23
clarify about crypto as a base image

glicht:
## Status
Ready

## Description
Add clarification about using the demisto/crypto as a base.


GalRabin (approved):
Looks good.

PR #2800, opened 2020-09-29
Hod/Configuration manager

hod-alpert:
## Status
Ready

## Related Issues
related: https://github.com/demisto/etc/issues/10273

## Related PRs 
https://github.com/demisto/dockerfiles/pull/2809

## Description
Using a new base for the powershell remoting image, no need for open ssh for thos one, only WSMan which is already installed

### DIFF
--- a/docker/pwsh-remoting/Dockerfile
+++ b/docker/pwsh-remoting/Dockerfile
@@ -1,9 +1,3 @@
-FROM demisto/powershell-ubuntu:7.0.1.9103
+FROM devdemisto/powershell-ubuntu:7.0.3.11992
 
-RUN mkdir /root/.ssh
-#Supressing the interactive host verification prompt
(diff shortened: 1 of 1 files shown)


docker/pwsh-remoting/Dockerfile
```diff
@@ -1,9 +1,3 @@
-FROM demisto/powershell-ubuntu:7.0.1.9103
+FROM devdemisto/powershell-ubuntu:7.0.3.11992
```
Itay4, line 1: we shouldnt use `devdemisto`

hod-alpert: Yes, it's only temporary until I'll push the new 7.0.3 version in this PR https://github.com/demisto/dockerfiles/pull/2809

PR #2828, opened 2020-10-01
ansible-runner for ansible 2.9.12

SergeBakharev:
## Status
In-Progress

## Related Issues
N/A

## Description
Required container image for my Automation Rising 2020 SOAR Hackathon submission: [1+1=3 Supercharging XSOAR with Ansible](https://github.com/demisto/content/pull/9197)

Two Python packages (`ansible`, and it's dependency `docutils`) do not pass verify_licenses.py. However I'm unclear if GPL is a approved OSI license type or not. it is not present in known_licenses.json nor approved_licenses.json. 

My submission does not link to the Ansible python libraries, so I do not believe the GPL Automatic Inheritance applies to my submission (if that was a concern), and even if it did I intend on making my hackathon submission available to the community. 

Using the python ansible module is merely a convenient way to get a specific version of Ansible install into the container that I have tested my work against. The Debian Buster ansible package is quite a few versions out of date from what I would like to use, but the Debian Bullseye is one patch level ahead. If use of the ansible python module is not allowed, I am happy to use the Debian ansible package as an alternative.


glicht:
Thanks for your submission. I am looking into this and will update the code as needed to get the build passing.

glicht (approved):

PR #2845, opened 2020-10-05
Build ML docker

yguezpa:
## Status
In Progress

## Related Issues
related: Create demisto ML python3 docker as part of dockerfile repo #28938

## Description
Create a new docker for ML tasks that we contains packages needed for Phishing (for now) and will be updated in the future with more packages


docker/ml/Pipfile
```diff
@@ -0,0 +1,12 @@
+[[source]]
+name = "pypi"
+url = "https://pypi.org/simple"
+verify_ssl = true
+
+[dev-packages]
+
+[packages]
```
glicht, line 8: all python packages should be defined here

docker/ml/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM demisto/python3-deb:3.8.5.11787
+
+COPY requirements.txt .
+
+RUN apt-get update && apt-get install -y    python3-dev g++
+
+RUN pip3 install spacy
```
glicht, line 7: you should not install packages via pip3 in the Dockerfile. This is done via the use of the Pipfile and Pipfile.lock.

glicht (changes requested):

yguezpa:
create a new PR

PR #2938, opened 2020-10-18
Add nuclei Docker image

rotemreiss:
## Status
In Progress - Docker image seems to work, but supporting integration is still WIP.

## Related Issues
None.

## Description
What is nuclei - https://github.com/projectdiscovery/nuclei .

> Nuclei is a fast tool for configurable targeted scanning based on templates offering massive extensibility and ease of use.
> 
> Nuclei is used to send requests across targets based on a template leading to zero false positives and providing effective scanning for known paths. Main use cases for nuclei are during initial reconnaissance phase to quickly check for low hanging fruits or CVEs across targets that are known and easily detectable. It uses retryablehttp-go library designed to handle various errors and retries in case of blocking by WAFs, this is also one of our core modules from custom-queries.


glicht:
@rotemreiss thanks for your contribution. Is this still "In Progress"?

rotemreiss:
Hi @glicht , that automation is still in progress, but those are two separate things. I think that you can merge this one ;)

glicht:
@rotemreiss what is the status with the integration?

rotemreiss:
Hi @glicht,
As already mentioned, you can merge it since it is working just fine.
We may want to add some improvements in the future, but we will make sure to have backward compatibility as every other plugin.

docker/nuclei/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM projectdiscovery/nuclei as builder
+
+FROM demisto/python3:3.8.6.12176
+
+COPY requirements.txt .
+
+RUN mkdir /app
+
+RUN apk --update add --no-cache --virtual .build-dependencies python3-dev build-base wget git \
+  && git clone https://github.com/projectdiscovery/nuclei-templates.git /app/nuclei-templates \
```
glicht, line 10: @rotemreiss Can we do here a shallow clone to keep the image size small? Also do you want to always work on master? Is master always stable or is there need to work on a specific tag?

glicht: Maybe also delete /app/nuclei-templates/.git after the clone?

rotemreiss: Actually, I will need to check it, since I think that on the latest version of Nuclei, they are already fetching the templates automatically and this clone can completely be removed.
I'll share an update asap.

glicht:
@rotemreiss any update on this?

docker/nuclei/Dockerfile
glicht: @rotemreiss you never got back on this. Is this still relevant?

PR #2939, opened 2020-10-18
CircleCI Udpate to python3.8 enviorment

GalRabin:
FYI - @yaakovi @reutshal 

## Status
Ready

## Related Issues
related: https://github.com/demisto/etc/issues/29577

## Description
Walrus and other python38 feature currently not supported in our content repo due to the fact that they fail when running lint in CircleCI, the next steps in order to support it after update the docker file is to replace it in config.yml and verify it on the nightly run.

Reason for failure: Mypy run locally and the interpreter we have locally in CircleCI is python37

### DIFF
--- a/docker/content-build/Dockerfile
+++ b/docker/content-build/Dockerfile
@@ -1,16 +1,57 @@
-FROM circleci/python:3.7.5-buster-node
+FROM cimg/base:2020.01
 
-RUN sudo apt-get update && \
-  sudo apt-get install -y --no-install-recommends python-dev libxml2-dev libxslt1-dev zlib1g-dev python-setuptools jq \
-  make build-essential libssl-dev libbz2-dev \
-  libreadline-dev libsqlite3-dev wget curl llvm libncurses5-dev libncursesw5-dev \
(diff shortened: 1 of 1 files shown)


docker/content-build/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM circleci/python:3.7.5-buster-node
+FROM circleci/3.8.6-buster-node-browsers
```
glicht, line 1: The version you put in seems wrong. It should be:
```suggestion
FROM circleci/python:3.8.6-buster-node
```

GalRabin: Right

glicht:
@GalRabin why all these changes just to upgrade the python version?

GalRabin:
> @GalRabin why all these changes just to upgrade the python version?

Due to the issue, I told you about on bootstrap which also reproduced in other environments (https://app.circleci.com/pipelines/github/demisto/content/35037/workflows/1960a872-087b-46b0-b9df-d423b54fb387/jobs/152604)

Link to issue: https://github.com/demisto/etc/issues/29637
We must find a solution to it.

PR #2940, opened 2020-10-18
boto3 python 2 image

glicht:
## Status
Ready

## Related Issues
related: https://github.com/demisto/etc/issues/29539

## Description
Python 2 based boto3 image for integrations which are still using py2 and boto3.


Itay4 (approved):
perhaps we should name it `boto3py2` so it will be aligned with `boto3py3` ?

glicht:
Better to leave the name to avoid using the old one.

PR #3002, opened 2020-10-25
Added docker file to cylance-protect-v2

abaumgarten:
## Status
Ready

## Related Issues
related: https://github.com/demisto/content/pull/9550

## Description
Added a docker file to pyjwt (did not have a docker file before) 
https://hub.docker.com/r/demisto/cylance_protect_v2/tags


glicht (changes requested):
@abaumgarten I don't see here anything specific for cylance. It seems that the only library it is using is pyjwt. 

Please create the docker image `demisto/pyjwt` which is also missing from the project and let's use this. 

There is no need for `virtualenv` and for `uuid` it is better to use the one provided by python: https://docs.python.org/2/library/uuid.html. Thus only needed pyjwt.

glicht (approved):
Looks good.

PR #3044, opened 2020-10-28
Py2-Exchangelib

Bargenish:
## Status
In progress 

## Related Issues
related: https://github.com/demisto/etc/issues/29245

## Description
Added a new python2 exchangelib(v 2.2.0) docker file in order to support the new `reply-mail` command.

## Note:
due to exchangelib limitation, we may delete these changes and stay with the old docker image


glicht:
Can you link the related issue in etc?

glicht (approved):

PR #3101, opened 2020-11-03
Sane pdf: non root 2

glicht:
## Status
Ready

## Related Issues
related: link to the issue

## Description
Change to chown also on /app dir for completenes.


glicht:
@dorsha I noticed that the previous PR was missing chown on /app. Changed this here and also using root user and not just group for completeness.

dorsha (approved):

PR #3127, opened 2020-11-10
updating docker for devo

roysagi:
## Status

## Related Issues
related: https://github.com/demisto/etc/issues/30376

## Description
updating devodsconnector to 2.2.1

### DIFF
--- a/docker/devo/Pipfile
+++ b/docker/devo/Pipfile
@@ -6,7 +6,7 @@ verify_ssl = true
 [dev-packages]
 
 [packages]
-devodsconnector = "==2.2.1"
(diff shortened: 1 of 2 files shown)


docker/devo/Pipfile
```diff
@@ -6,7 +6,7 @@ verify_ssl = true
 [dev-packages]
 
 [packages]
-devodsconnector = "*"
+devodsconnector = "==2.2.1"
```
glicht, line 9: It is enough to update just in the Pipfile.lock. You should leave this as `*`

glicht (approved):

PR #3249, opened 2020-11-23
adding dhs-feed docker img

esharf:
## Status
Ready.

## Related Issues
related: https://github.com/demisto/content/pull/9960

## Description
adding this because I need both packages xmltodict to pars taxii response and OpenSSL to test the creds (for better test command)


docker/dhs-feed/Dockerfile
```diff
@@ -0,0 +1,13 @@
+
+FROM demisto/python3-deb:3.8.6.13363
```
glicht, line 2: Use as a base: `demisto/crypto:1.0.0.12979` which already takes care of importing crypto and is alpine based. See for example: https://github.com/demisto/dockerfiles/tree/master/docker/teams

glicht (changes requested):
Let give this docker image a generic name. You can call it `xml-feed` as there is nothing specific to dhs.

glicht (approved):

PR #3280, opened 2020-11-29
luminate docker image

glicht:
## Status
Ready

## Related Issues
related: https://github.com/demisto/etc/issues/14971

## Description


docker/luminate/.gitignore
```diff
@@ -0,0 +1 @@
+requirements.txt
```
anara123, line 1: why to ignore requirements file?

glicht: we do this for all images as the file is generated from the Pipfile and if for some reason the build fails we don't want the dev to checkin this file by mistake.

anara123 (approved):

PR #3308, opened 2020-12-04
GreyNoise python lib

crestdatasystems:
## Status
Ready

## Description
GreyNoise lib for Demisto/Cortex XSOAR integration


glicht:
@crestdatasystems the greynoise lib imports dicttoxml library which is GPL. See: https://github.com/quandyfactory/dicttoxml/blob/master/LICENCE.txt . We need to understand how this library will be used before merging. Please reach out on slack (DFIR) or via email.

crestsantoshgiri:
@glicht @bakatzir 
 - updated PR with a resolution of dicttoxml library

glicht (approved):

PR #3435, opened 2020-12-17
Content build - Install build requirements

GalRabin:
## Status
In Progress

## Description
Install build pre-requirements


glicht:
@GalRabin we had this in the past and decided that it is easier to manage the build dependencies in the content repo. Also, since we are caching the venv it doesn't provide much value regarding build setup.

PR #3592, opened 2020-12-31
add ODBC driver to genericsql

adi88d:
## Status
Ready

## Related Issues
related: https://github.com/demisto/etc/issues/30431

## Description
add ODBC driver to genericsql

### DIFF
--- a/docker/genericsql/Dockerfile
+++ b/docker/genericsql/Dockerfile
@@ -71,3 +71,6 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
   python3-dev \
   unixodbc-dev \
 && rm -rf /var/lib/apt/lists/*
+
+# Handling the issue described here https://github.com/mkleehammer/pyodbc/issues/610#issuecomment-534920201 by using TLS v1
+RUN sed -i -e 's/MinProtocol = TLSv1.2/MinProtocol = TLSv1.0/' -e 's/CipherString = DEFAULT@SECLEVEL=2/CipherString = DEFAULT@SECLEVEL=1/' /etc/ssl/openssl.cnf
--- a/docker/genericsql/Dockerfile
+++ b/docker/genericsql/Dockerfile
@@ -52,12 +52,7 @@ RUN apt-get update \
 && curl https://packages.microsoft.com/config/debian/10/prod.list > /etc/apt/sources.list.d/mssql-release.list \
 && apt-get update \
 && ACCEPT_EULA=Y apt-get install msodbcsql17 \
-&& ACCEPT_EULA=Y apt-get install mssql-tools \
-&& echo 'export PATH="$PATH:/opt/mssql-tools/bin"' >> ~/.bash_profile \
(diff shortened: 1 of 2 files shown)


glicht:
@adi88d build is failing as curl is missing. you can use wget instead to download and then run the file.

docker/genericsql/Dockerfile
```diff
@@ -45,6 +45,19 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
...
 
+# Install the Microsoft ODBC driver for SQL Server (Linux)
+RUN apt-get update \
+&& apt-get install -y --allow-downgrades gpgv=2.2.12-1+deb10u1 gnupg2 curl apt-utils \
+&& curl https://packages.microsoft.com/keys/microsoft.asc | apt-key add - \
+&& curl https://packages.microsoft.com/config/debian/10/prod.list > /etc/apt/sources.list.d/mssql-release.list \
+&& apt-get update \
+&& ACCEPT_EULA=Y apt-get install msodbcsql17 \
+&& ACCEPT_EULA=Y apt-get install mssql-tools \
+&& echo 'export PATH="$PATH:/opt/mssql-tools/bin"' >> ~/.bash_profile \
```
glicht, line 56: Not sure what this is for. But it probably doesn't have the effect you want as we don't load the bash profile when running integrations.

glicht:
@hod-alpert see my comment

docker/genericsql/verify.py
```diff
@@ -1,15 +1,14 @@
 import sqlalchemy
-import pymysql
 import pyodbc
-import psycopg2
```
glicht, line 4: Why did you remove this? We use this to make sure the library is imported correctly

docker/genericsql/verify.py
```diff
@@ -1,15 +1,14 @@
 import sqlalchemy
-import pymysql
```
glicht, line 2: Why did you remove this? We use this to make sure the library is imported correctly

glicht (commented):
@hod-alpert see my comment about verify.py script

docker/genericsql/Dockerfile
```diff
@@ -63,3 +71,6 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
   python3-dev \
   unixodbc-dev \
 && rm -rf /var/lib/apt/lists/*
+
+# Handling the issue described here https://github.com/mkleehammer/pyodbc/issues/610#issuecomment-534920201 by using TLS v1
+RUN sed -i -e 's/MinProtocol = TLSv1.2/MinProtocol = TLSv1.0/' -e 's/CipherString = DEFAULT@SECLEVEL=2/CipherString = DEFAULT@SECLEVEL=1/' /etc/ssl/openssl.cnf
```
glicht, line 76: Excellent. I was thinking of the same. Please use the same sed we use at other images. So it will be uniform. See:

https://github.com/demisto/dockerfiles/blob/master/docker/py2-exchangelib/Dockerfile#L16

glicht (approved):

PR #3648, opened 2021-01-06
adding stopwords to docker file

lperryp:
## Status

## Related Issues
related: link to the issue

## Description
adding stopwords to sklearn docker file


yguezpa (approved):
@lperryp Don't you use ml docker instead of sklearn btw ?

lperryp:
@yguezpa yes, but this is much lighter so I prefer to use this one when possible

PR #3687, opened 2021-01-11
create coco-loco docker image

barchen1:
## Status
Ready

## Description
create opencv docker image


glicht:
@barchen1 let's name the image opencv

Did you try to build it with alpine based image?

barchen1:
> @barchen1 let's name the image opencv
> 
> Did you try to build it with alpine based image?

I did, but I had some issues with it. 
https://stackoverflow.com/questions/50950588/trouble-installing-opencv-in-docker-container-using-pip

.dependabot/config.yml
```diff
@@ -1180,3 +1180,59 @@ update_configs:
         dependency_name: demisto/*
         update_type: all
 
+  - package_manager: python
```
glicht, line 1183: not clear why you have this entry. Maybe something local.

.dependabot/config.yml
```diff
@@ -1180,3 +1180,59 @@ update_configs:
...
+        update_type: all
+
+  - package_manager: python
+    directory: /docker/loco-loco
+    update_schedule: live
+    automerged_updates:
+    - match:
+        update_type: semver:minor
+  - package_manager: docker
+    directory: /docker/loco-loco
```
glicht, line 1204: remove the old loco stuff.

glicht:
@barchen1 let's create a base image opencv. Then the yolo specific files we will use in a new image which is based on opencvc.

docker/opencv/Dockerfile
```diff
@@ -0,0 +1,15 @@
+
+FROM demisto/python3-deb:3.9.1.14969
+
+COPY requirements.txt .
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+  gcc \
+  python-dev \
+&& apt install python3 python3-distutils wget libglib2.0-0 libsm6 libxext6 libgl1-mesa-glx libxrender-dev -y \
```
glicht, line 9: You should install also libxrender (not only the `-dev`). and in the purge section remove the `libxrender-dev` install.

Aslo I am not clear why you are installing `python3 python3-distutils` . I don't think this should be needed.

barchen1: done

glicht (approved):

PR #3978, opened 2021-02-07
PWSH-Exchange: Upgraded base docker and added SelfSignedCertificate

jochman:
## Status
Ready

## Description
Added the SelfSignedCertificate 3rd party pack.
Upgraded docker to ubuntu 7.1.1

### DIFF
--- a/docker/pwsh-exchange/build.conf
+++ b/docker/pwsh-exchange/build.conf
@@ -1,1 +1,1 @@
-version=7.1.1
+version=1.0.0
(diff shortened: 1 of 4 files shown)


docker/pwsh-exchange/build.conf
```diff
@@ -1 +1 @@
-version=1.0.0
+version=7.1.1
```
glicht, line 1: No need for this bump. We only bump the version to the base images.

jochman: I thought that we should match the docker version to the pwsh version.

glicht (approved):

PR #4140, opened 2021-02-15
New docker image - OpenCTI-v4

daryakoval:
## Status
Ready

## Description
New docker image - OpenCTI-v4


.dependabot/config.yml
```diff
@@ -1953,42 +1953,43 @@ update_configs:
     - match:
         dependency_name: demisto/*
         update_type: all
-        
+
   - package_manager: python
-    directory: /docker/opencti-v4.1.2
+    directory: /docker/xgboost
```
glicht, line 1958: @daryakoval something is weird with the changes you are putting in the dependabot config file. Can you please check.

glicht:
@darkushin @ShahafBenYakir we have the following image: https://github.com/demisto/dockerfiles/tree/master/docker/opencti

Why do we need this one?

daryakoval:
@glicht - this https://github.com/demisto/dockerfiles/tree/master/docker/opencti working with OpenCTI integration, 
and the new docker works with OPENCTI v2. 
we cant use the same docker since the API changed and it breaks backward compatibility.

.dependabot/config.yml
daryakoval: git looks at this in strange way.. I just deleted one docker and added another..

daryakoval: Instead of deleting previous docker image and add new one... it shifts all docker names down

ShahafBenYakir (approved):

PR #4217, opened 2021-02-21
xsoar-tools - bump demisto-sdk version to 1.3.0

Itay4:
## Status
Ready

## Description
bump demisto-sdk version


glicht:
Note that you are bumping here also other deps such as mypy.

glicht:
We probably are better using an exact version in Pipfile for the deps we use in the build (at least the linters). See: https://github.com/demisto/content/blob/master/dev-requirements-py3.txt

glicht (approved):

Itay4:
@glicht please see my last changes
the common server scripts are required for the lint to avoid false positives about missing funcs

glicht:
> @glicht please see my last changes
> the common server scripts are required for the lint to avoid false positives about missing funcs

I think it is better to download just the relevant files using wget and the raw url. For example: https://raw.githubusercontent.com/demisto/content/master/Packs/Base/Scripts/CommonServerPython/CommonServerPython.py

Itay4:
@glicht modified to use wget

glicht (approved):

PR #4621, opened 2021-03-29
Adding support for pytorch + new oob binary model

lperryp:
## Description
Adding support for pytorch _ adding model binary to docker
## Status

## Related Issues
related: link to the issue


docker/ml/Dockerfile
```diff
@@ -1,14 +1,18 @@
...
+RUN apt-get update && apt-get install -y --no-install-recommends \
+  gcc \
+  python3-dev gfortran wget git g++ pkg-config libhdf5-dev \
+&& ln -s /usr/include/locale.h /usr/include/xlocale.h \
 && pip install --no-cache-dir -r requirements.txt \
 && apt-get purge -y --auto-remove \
-  g++ \
-  python3-dev
+  gcc \
+  python3-dev gfortran wget git g++ pkg-config libhdf5-dev
```
yguezpa, line 11: Why are you installing **gcc python3-dev gfortran wget git g++ pkg-config libhdf5-dev** twice ? 
I think you can install them and then remove them  no ? If not you can install them once after **apt-get purge -y --auto-remove**

docker/ml/Dockerfile
```diff
@@ -1,14 +1,18 @@
...
 COPY demisto_ml-0.2.tar.gz .
 RUN pip install demisto_ml-0.2.tar.gz
 RUN rm -rf /var/lib/apt/lists/*
 RUN python -m spacy download en_core_web_sm
-
-
+RUN python -c "import nltk; nltk.download('wordnet'); nltk.download('stopwords'); nltk.download('punkt')"
+COPY encrypted_model.b ./var
+COPY oob_evaluation.txt ./var
```
yguezpa, line 18: Do we want to copy the model in any case in the docker ? even if this docker is used for another purpose ?

docker/ml/verify.py
```diff
@@ -1,5 +1,4 @@
-import tensorflow
-import keras
```
yguezpa, line 2: You install Keras in the docker so you have to make sure it import it correctly. 
Btw the model I am building for system diagnostic uses Keras (I don't like Keras but since I believed your model use Keras I used it to build it). Not sure if this model is going somewhere but let's make sure this docker can use keras and pytorch. It will be hard to add it afterward

yguezpa (changes requested):

docker/ml/Dockerfile
```diff
@@ -1,14 +1,16 @@
...
-&& apt-get purge -y --auto-remove \
-  g++ \
-  python3-dev
+
+RUN apt-get update \
+&& apt-get install -y --no-install-recommends \
+  gcc \
+  python3-dev gfortran wget git g++ pkg-config libhdf5-dev \
+&& ln -s /usr/include/locale.h /usr/include/xlocale.h \
+&& pip install --no-cache-dir -r requirements.txt
```
yguezpa, line 9: You need to remove all the dev packages after the pip installation

yguezpa (changes requested):

yguezpa (approved):

lperryp:
@glicht  - can you have also a look please? made some changes with the packages, and docker size increased significantly

PR #4873, opened 2021-04-11
Created image to work on for gitlab content build.

DinaMeylakh:
## Status
fixes: https://github.com/demisto/etc/issues/35419


docker/gitlab-content-ci/Pipfile
```diff
@@ -0,0 +1,55 @@
+[[source]]
+name = "pypi"
+url = "https://pypi.org/simple"
+verify_ssl = true
+
+[dev-packages]
+
+[packages]
```
avidan-H, line 8: Do we need all these packages? Don't we use the virtual environment in the `venv` directory for most python script executions?

docker/gitlab-content-ci/Dockerfile
```diff
@@ -24,5 +24,8 @@ RUN  apt-get update -y \
 # Currently sh is linked to dash, linking it to /bin/bash instead
 RUN rm /bin/sh && ln -s /bin/bash /bin/sh
 
-# Creating a user with home directory that matces the home directory of github actions, and give the user sudo access
-#RUN groupadd --gid 1000 github && mkdir -p /github/ && useradd -m -d /github/home --uid 1000 --gid root --shell /bin/bash github
+# Installing nvm for the private build
+RUN \
+    curl https://raw.githubusercontent.com/creationix/nvm/master/install.sh | bash && \
```
glicht, line 29: Don't install from master. Use a tag. Such as: `https://raw.githubusercontent.com/nvm-sh/nvm/v0.38.0/install.sh`

docker/gitlab-content-ci/Pipfile
DinaMeylakh: As discussed, this is done to eliminate the usage of build-requirements.txt file and the packages here are the ones needed for the build environment to run.

docker/gitlab-content-ci/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+
+RUN apt-get install -y curl jq unzip openssh-client gcc libffi-dev libssl-dev git zip netcat wget python2 python-pip
+
+# Currently sh is linked to dash, linking it to /bin/bash instead
+RUN rm /bin/sh && ln -s /bin/bash /bin/sh
+
+# Installing nvm, npm and node.js
+ENV NODE_VERSION 12.18.4
+RUN \
+    curl https://raw.githubusercontent.com/creationix/nvm/v0.38.0/install.sh | bash \
```
avidan-H, line 13: Why is this,
`https://raw.githubusercontent.com/creationix/nvm/v0.38.0/install.sh`
and not,
`https://raw.githubusercontent.com/nvm-sh/nvm/v0.38.0/install.sh`
?

docker/gitlab-content-ci/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+
+ENV NVM_DIR /root/.nvm
+ENV NODE_PATH $NVM_DIR/versions/node/v$NODE_VERSION/lib/node_modules
+ENV PATH $NVM_DIR/versions/node/v$NODE_VERSION/bin:$PATH
+
+COPY requirements.txt .
+
+RUN pip install --no-cache-dir -r requirements.txt
+
+# Instaling the gsutil
```
avidan-H, line 29:
```suggestion
# Installing the gsutil
```

docker/gitlab-content-ci/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+    curl https://raw.githubusercontent.com/creationix/nvm/v0.38.0/install.sh | bash \
+    && export NVM_DIR="/root/.nvm" \
+    && export NODE_VERSION="12.18.4" \
+    && . $NVM_DIR/nvm.sh \
+    && nvm install $NODE_VERSION \
+    && nvm alias default $NODE_VERSION \
+    && nvm use default
+
+ENV NVM_DIR /root/.nvm
+ENV NODE_PATH $NVM_DIR/versions/node/v$NODE_VERSION/lib/node_modules
```
avidan-H, line 22: Why are you explicitly setting this? Isn't the benefit of using nvm the flexibility? Shouldn't this stuff be set during job execution depending on what job and what repo is using this image?

avidan-H (changes requested):

docker/gitlab-content-ci/Dockerfile
DinaMeylakh: This is set for the default npm usage.
you still have to `source $NVM_DIR/nvm.sh` in the container to use nvm at all, so the NVM_DIR should be defined as installed.
The rest is for the default npm usage which will only work if you explicitly add `nvm use default` to your container.
It seemed more logical to add a default node version to the nvm instead of installing it every time in the build.

docker/gitlab-content-ci/Dockerfile
DinaMeylakh: good catch

docker/gitlab-content-ci/Dockerfile
avidan-H: > This is set for the default npm usage.
you still have to source $NVM_DIR/nvm.sh in the container to use nvm at all, so the NVM_DIR should be defined as installed.
I have no issue with setting an nvm root directory environment variable. I don't understand why you need to set the `NODE_PATH` variable.

avidan-H (approved):

PR #4874, opened 2021-04-12
Contrib GreyNoise update to version 0.8.0

crestdatasystems:
## Status
**Ready**

## Description
This is an abstract python library built on top of the GreyNoise service. It is preferred that users use this library when implementing integrations or plan to use GreyNoise within their code. The library includes a small client to interact with the API.


crestdatasystems:
Can we please merge this PR as GreyNoise new change is live.
cc: @anara123

roysagi:
@glicht  - I think we need your approval here

glicht (approved):

PR #4927, opened 2021-04-19
Support non root permissions for ML docker 

lperryp:
## Status

## Related Issues
related: https://github.com/demisto/etc/issues/35834

## Description
Adding support for running nltk and matplotlib packages without root permissions


lperryp:
@glicht  - can you please check this? we're having issues when `docker.run.internal.asuser==true`, and python packages can't access their resources on the docker, this solution seems like the right approach?

erezh31 (approved):

docker/ml/Dockerfile
```diff
@@ -12,7 +12,16 @@ RUN apt-get update \
...
+
+USER ml
+
 RUN python -m spacy download en_core_web_sm
-RUN python -c "import nltk; nltk.download('stopwords'); nltk.download('punkt')"
-COPY encrypted_model.b ./var
-COPY oob_evaluation.txt ./var
+RUN python -c "import nltk; nltk.download('stopwords', download_dir='/ml/nltk_data'); nltk.download('punkt',download_dir='/ml/nltk_data')"
+ENV NLTK_DATA='/ml/nltk_data'
+ENV MPLCONFIGDIR='ml/matplotlib'
```
glicht, line 25: Don't you need a `/`
```suggestion
ENV MPLCONFIGDIR='/ml/matplotlib'
```

lperryp: thanks

glicht (changes requested):
The part about adding the `ml` user can be removed. We run with a dedicated user. I would remove the following:
```
WORKDIR /ml
RUN useradd -m ml
RUN chown -R ml:ml /ml

USER ml
```

**after** the download is done do the following:

```
RUN chown -R demisto:demisto /ml && chmod -R 775 /ml
```

glicht (approved):

PR #5040, opened 2021-04-29
adding new dev dependency in Carbon Black API image

ilappe:
## Status
Ready

## Related Issues
related: demisto/etc#31243

## Description
Adding new dev dependency on Carbon Black specific module

### DIFF
--- a/docker/carbon-black-cloud/Pipfile
+++ b/docker/carbon-black-cloud/Pipfile
@@ -7,6 +7,7 @@ name = "pypi"
 carbon-black-cloud-sdk = "*"
 
 [dev-packages]
+cbapi = "*"
 
 [requires]
 python_version = "3.9"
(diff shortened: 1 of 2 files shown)


esharf (approved):

docker/carbon-black-cloud/Pipfile
```diff
@@ -7,6 +7,7 @@ name = "pypi"
 carbon-black-cloud-sdk = "*"
 
 [dev-packages]
+cbapi = "*"
```
glicht, line 10: @esharf @ilappe I think this was added at the wrong place. Under dev-packages instead of packages. Please check.

ilappe: already fixed in #5041 
Thanks

PR #5365, opened 2021-05-27
Sane doc reports -> public

agamm:
## Status
In Progress

demisto/sane-doc-reports#87

## Related Issues
Related: demisto/etc#37148

## Description
Make changes to add the public repo of sane-doc-reports & update a fix.


glicht (changes requested):
Please change to use pipenv instead of `git clone`. Do the following in the dir `docker/sane-doc-reports`:
```
pipenv install  -e 'git+https://github.com/demisto/sane-doc-reports.git@master#egg=sane-doc-reports'
```
This will modify the Pipenv and Pipenv.lock files to install from github. Furthermore, the lock file will point to a specific commit. Then remove the `git clone` part in Dockerfile.

glicht:
@agamm see my commit

agamm:
> @agamm see my commit

Looks good, but it still doesn't use pipenv, right?

Plus, shouldn't the base image for python include: `gcc` and `python3-dev` ?

agamm:
Let's merge?

glicht (approved):

PR #6018, opened 2021-07-15
Fix for Powershell host trust issue

amshamah419:
## Status
In Progress

## Related Issues
Related: https://github.com/demisto/etc/issues/38964

## Description
The module installation was missing `-Scope AllUsers` which would prevent it from being installed with the proper permissions required for host users. I added this and the test in the issue mentioned is resolved.


glicht (approved):
Good catch.

PR #6293, opened 2021-08-16
add golangcli-lint to gitlabci

asafshen:
## Status
Ready 

Description
add golangcli-lint to gitlabci image


asafshen:
I found another image that has what I need in order to use golangcli-lint, closing this one

PR #6300, opened 2021-08-27
Gitlab ci marketplace

yardensade:
mostly due to ubuntu 18.04 needed


docker/gitlab-ci-marketplace/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+    ca-certificates gnupg bzip2 openssh-client libxml2-dev \
+    libxslt1-dev libxslt-dev zlib1g-dev libxmlsec1 xmlsec1 \
+    libxml2-dev libxmlsec1-dev libxmlsec1-openssl pcregrep
+
+RUN curl -fsSL -O https://download.docker.com/linux/static/stable/x86_64/docker-19.03.9.tgz && \
+    tar xf docker-19.03.9.tgz && \
+    mv docker/* /usr/bin/ && \
+    rm -rf docker-19.03.9.tgz docker
+
+RUN groupadd circleci && groupadd docker \
```
asafshen, line 18: why `circleci`?

asafshen (commented):
@yardensade overall looks good, is this copied from another images?

@glicht I would appreciate your look here as well 

also need to make sure all relevant places places will use the same image (marketplace gateway and marketplace auth proxy)

yardensade:
This is the same docker image you used initially for the CI except I removed some parts that we did not need and changed to ubuntu 18 which we the same OS we used in github.

I don't mind cleaning it more but this takes time and I first want to test that the deploy script works, then we can easily remove things knowing it did not ruin anything. The MP ci repos that don't use postgres are less in priority to work with this one, we can change them afterwards when we are done with subs/reviews.

asafshen:
> I don't mind cleaning it more but this takes time and I first want to test that the deploy script works, then we can easily remove things knowing it did not ruin anything. The MP ci repos that don't use postgres are less in priority to work with this one, we can change them afterwards when we are done with subs/reviews.


I agree with the approach of having a dedicate image. no need to clean, I completely  agree it doesn't worth the time ATM

I am just wondering if we can use the same docker image as a base layer, (so if we will have to fix/add something to the original image it will reflect here as well)

approving anyhow, can be changed later

asafshen (approved):

yardensade:
We should be able to, I will put it on my list to check it out, should be fairly simple and quick, I agree that it is best to have all MP projects on a single image if we already took the time to make one.

PR #6321, opened 2021-09-29
Enhance gitlab-content-ci

DeanArbel:
## Status
In Progress

## Related Issues
Related: link to the issue

## Description
Add goenv

### DIFF
--- a/docker/gitlab-content-ci/Dockerfile
+++ b/docker/gitlab-content-ci/Dockerfile
@@ -34,4 +34,10 @@ RUN  apt-get update -y \
   && gcloud --version \
   && gsutil --version
 
-
+# Installing goenv
+RUN git clone https://github.com/syndbg/goenv.git ~/.goenv \
+    && export GOENV_ROOT="$HOME/.goenv" \
+    && export PATH="$GOENV_ROOT/bin:$PATH" \
+    && eval "$(goenv init -)"
+    && export PATH="$GOROOT/bin:$PATH" \
+    && export PATH="$PATH:$GOPATH/bin"


avidan-H:
@DinaMeylakh can you take a look, please? You're the owner here. :octocat:

tkatzir (changes requested):
Merge `RUN` commands to one.
e.g. the `apt-get` (update and install)

docker/gitlab-content-ci/Dockerfile
```diff
@@ -34,4 +34,15 @@ RUN  apt-get update -y \
...
+  && echo 'export GOENV_ROOT="$HOME/.goenv"' >> ~/.bashrc \
+  && echo 'export PATH="$GOENV_ROOT/bin:$PATH"' >> ~/.bashrc \
+  && echo 'eval "$(goenv init -)"' >> ~/.bashrc \
+  && echo 'export PATH="$GOROOT/bin:$PATH"' >> ~/.bashrc \
+  && echo 'export PATH="$PATH:$GOPATH/bin"' >> ~/.bashrc \
+  && source ~/.bashrc \
+  && goenv install 1.13.8 \
+  && goenv global 1.13.8 \
+  && source ~/.bashrc \
+  && curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.40.1
```
DinaMeylakh, line 48: better not to install stuff from the master branch of another repo, if there's a branch corresponding to a version then we should use it.

docker/gitlab-content-ci/Dockerfile
```diff
@@ -34,4 +34,15 @@ RUN  apt-get update -y \
...
   && gsutil --version
 
-
+# Installing goenv
+RUN git clone https://github.com/syndbg/goenv.git ~/.goenv \
+  && echo 'export GOENV_ROOT="$HOME/.goenv"' >> ~/.bashrc \
+  && echo 'export PATH="$GOENV_ROOT/bin:$PATH"' >> ~/.bashrc \
+  && echo 'eval "$(goenv init -)"' >> ~/.bashrc \
+  && echo 'export PATH="$GOROOT/bin:$PATH"' >> ~/.bashrc \
+  && echo 'export PATH="$PATH:$GOPATH/bin"' >> ~/.bashrc \
```
DinaMeylakh, line 43: Did you verify the environment variables you exported work as expected? Sometimes exporting the .bashrc doesn't propagate to gitlab and its better to use `ENV` when needed

DinaMeylakh (changes requested):
Looks good

docker/gitlab-content-ci/Dockerfile
DeanArbel: There is none - https://github.com/golangci/golangci-lint/branches/all
This is the recommended installation, see https://golangci-lint.run/usage/install/#other-ci

docker/gitlab-content-ci/Dockerfile
DeanArbel: Yeah - otherwise `goenv install 1.13.8` would fail.
Using the ENV variable is a good idea, as it'll ensure it persists, however since goenv updates need to run after the `goenv init` script, so it's recommended to put in the `.bashrc` file - the caveat for that is that it also requires to run it in the image itself.
The same goes for `nvm` env.

tkatzir (approved):

PR #6370, opened 2021-11-18
Handle Dockerfile.ironbank creation when requirements.txt is missing

adi88d:
## Status
Ready

## Related Issues
Related: https://github.com/demisto/etc/issues/43891

## Description
When the docker doesn't have `requirements.txt` we need to treat the `Dockerfile.ironbank` creation differently and remove those parts :
`COPY requirements.txt .`

- `RUN mkdir ./pip-pkgs`

- `COPY *.* ./pip-pkgs/`

- `RUN dnf install -y --nodocs python{0}-devel gcc gcc-c++ make wget git &&
        pip install --no-cache-dir --no-index --find-links ./pip-pkgs/ -r requirements.txt &&
        dnf remove -y python{0}-devel gcc gcc-c++ make wget git platform-python-pip &&
        dnf clean all && 
        rm -rf /var/cache/dnf &&
        rm -rf ./pip-pkgs`
        
the file `Dockerfile.ironbank` that created will be upload to repo1 in Ironbank project.


ironbank/build_ironbank_docker.sh
```diff
@@ -97,20 +97,23 @@ function build_dockerfile {
...
+  elif [[ -f $REQUIREMENTS ]]; then
+    python ./ironbank/build_dockerfile.py --docker_image_dir $1 --output_path $OUTPUT_PATH --requirements_file_exists true
   else
-    python ./ironbank/build_dockerfile.py --docker_image_dir $1 --output_path $OUTPUT_PATH
+    python ./ironbank/build_dockerfile.py --docker_image_dir $1 --output_path $OUTPUT_PATH --requirements_file_exists false
+  fi
+
+  if [[ ! -f $REQUIREMENTS ]] && [[ ! -f $1/Dockerfile.ironbank ]]; then
+    echo "docker_packages_metadata.txt is missing in this docker, please create Dockerfile.ironbank, aborting..."
+    return 1;
```
wissamg, line 116: No need for this condition, is not it?

adi88d: I keep this condition just in case we will change the flow in the future, so we will have one more validation

wissamg (approved):

PR #6371, opened 2021-11-21
Introducing github-native dependabot

ilaner:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/40952 

## Description
This PR introduces support for Github-native dependabot.
The old dependabot config file was migrated to the new syntax.
We don't use dependabot for docker images.


.github/v1tov2.py
```diff
@@ -0,0 +1,17 @@
+import yaml
+
+with open('/Users/ierukhimovic/dev/demisto/dockerfiles/.dependabot/config.yml') as f:
```
glicht, line 3: I think this is a local file you don't need to commiit

wissamg (approved):

PR #6572, opened 2021-12-01
Bump sane reports pdf idanizi

idanizi-panw:
## Status
Ready

## Description
Update sane report image


idanizi-panw:
will open another

PR #6678, opened 2021-12-14
Merge ml with nltk and dl-languages

agamm:
## Status
In progress

Related PR: --

## Related Issues
Related: https://github.com/demisto/etc/issues/35566#issue-852236190

## Description
Trying to merge both images into demisto/ml

Note:
Image will increase in 1.34GB

### DIFF
--- a/docker/ml/Dockerfile
+++ b/docker/ml/Dockerfile
@@ -21,8 +21,13 @@ RUN rm -rf /var/lib/apt/lists/*
 RUN python -m spacy download en_core_web_sm
 RUN mkdir /ml
 RUN python -c "import nltk; nltk.download('stopwords', download_dir='/ml/nltk_data'); nltk.download('punkt',download_dir='/ml/nltk_data')"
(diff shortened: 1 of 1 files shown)


lperryp:
Hi ! 

1. There's some duplication after change, we already download nltk in a different way in the dockerfile
2.  We currently don't need `dl-languages` (it was an old attempt to support multiple languages, which we now offer a different way for this which doesn't require this docker), and from what I see `demsito:nltk` is used by only one old test script, so they're not really needed

agamm:
> Hi !
> 
> 1. There's some duplication after change, we already download nltk in a different way in the dockerfile
> 2. We currently don't need `dl-languages` (it was an old attempt to support multiple languages, which we now offer a different way for this which doesn't require this docker), and from what I see `demsito:nltk` is used by only one old test script, so they're not really needed

1. Yeah I noticed, I just wanted the directories to be the same.
2. Ok
@wissamg What do you think? it seems like we don't need to merge. We can change the docker image in the content YAMLs.

PR #6684, opened 2021-12-15
Dependabot enhancements 

ilaner:
## Status
Ready

## Description
- Adding docker updates
- Adding and verifying docker updates
- handle case when update-type is unknown (using regex)
- add @wissamg as reviewer and add labels to the PRs.


.github/workflows/dependabot_auto_merge.yml
```diff
@@ -16,11 +16,22 @@ jobs:
...
-        if: ${{contains(steps.metadata.outputs.dependency-names, 'demisto/*') || steps.metadata.outputs.update-type == 'version-update:semver-patch' || steps.metadata.outputs.update-type == 'version-update:semver-minor'}}
+        if: ${{contains(steps.metadata.outputs.dependency-names, 'demisto/*') || steps.metadata.outputs.update-type == 'version-update:semver-patch' || steps.metadata.outputs.update-type == 'version-update:semver-minor' || steps.metadata.outputs.update-type == null}}
         env:
           PR_URL: ${{github.event.pull_request.html_url}}
           GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
+          PULL_TITLE: ${{ github.event.pull_request.title }}
         run: |
-          echo "Aprovving and merging"
+          if [[ -z steps.metadata.outputs.update-type ]]; then
+            minor_regex="^build\(deps(-dev)*\): bump .+ from ([[:digit:]]+)\.([[:digit:]]+)\.([[:digit:]]+) to ([[:digit:]]+)\.([[:digit:]]+)\.([[:digit:]]+)"
```
wissamg, line 26: Before you merge I think it is better to have $ sign by the end of the regex pattern to have full match

wissamg (approved):

.github/workflows/dependabot_auto_merge.yml
ilaner: It's not a full match because we missing the path of the docker

PR #6743, opened 2021-12-28
bumping up to python 3.9.9 for python3[-deb] and in the alpine case bumping to alpine 3.15

wissamg:
## Status
Ready

## Description
bumping up to python 3.9.9 for python3[-deb] and in the alpine case bumping to alpine 3.15


glicht:
What about py 2: https://github.com/demisto/dockerfiles/blob/master/docker/python/Dockerfile ? I think we can do this in this PR too.

glicht (approved):

PR #6791, opened 2022-01-05
Reduce images

agamm:
## Status
Ready

## Related Content Pull Request
Related PR: https://github.com/demisto/dockerfiles/pull/6685

## Related Issues
Related: https://github.com/demisto/etc/issues/44196

## Description
Reducing the smallest alpine python3 images we have into one 100mb limit image.
Which includes:
```
# Base size: 67mb
# Images:
# cloudshare, intezer, emoji, exifread, stomp.py, geopy, dns-python, 72mb
# algorithmia, asana, feed-parser-html, imap, pyjarm, python_zipfile 81mb
# paho-mqtt, python3-arrow, vertica, sendgrid, circleci, archer  85mb
# tweepy, tld, feed-parser, langdetect, gql, minio, ldap, pymongo 100mb
# cyjax, zabbix, jmespath, xlrd-py3, elasticsearch, aiohttp, stix2 120mb
```

PR #6814, opened 2022-01-06
Reduce images x2

agamm:
## Status
Ready

https://github.com/demisto/dockerfiles/pull/6791

## Description
After trying to use the image in content I found that some pip imports aren't used in the content at all, removed them.
Graphql is used also instead of gql which isn't used.

Ps it is 116mb now (just realized now we merged the 120mb version instead of the 100mb, no worries, more work done).


docker/py3-tools/Pipfile
```diff
@@ -21,20 +20,18 @@ imapclient = "*"
...
-circleci = "*"
 rsa-archer = "*"
 tweepy = "*"
 tld = "*"
 validate-email = "*"
 requests = "*"
 markdownify = "*"
 langdetect = "*"
-gql = "*"
+gql = "==3.0.0rc0"
```
wissamg, line 34: we should be careful with rc,  for now as for test propose we can have it

wissamg (approved):

docker/py3-tools/Pipfile
agamm: I chancged to the only used content usage of the graphql image (gql image isn't used) and graphql image uses the rc version. What do you say?

wissamg:
> 

I do not think that this will work, the integration should be configured to work insecure (no ssl verification)

agamm:
1 - was fixed already with the memory (in content repo conf.json, right?).
2 - fails in the nightly too

@wissamg  So can we merge?

docker/py3-tools/Dockerfile
```diff
@@ -7,14 +7,14 @@ COPY requirements.txt .
 RUN apk --update add --no-cache --virtual .build-dependencies python3-dev build-base wget git \
   && pip install --no-cache-dir -r requirements.txt \
   && apk del .build-dependencies
```
wissamg, line 9: missing a backslash by the end of the line or merge both lines 9 and 10

wissamg:
> 

Yes just fix the Dockerfile merge line 9 and 10 or add backslash by the end of line 9

agamm:
> > 
> 
> Yes just fix the Dockerfile merge line 9 and 10 or add backslash by the end of line 9

Done.

agamm:
Update all the images in the content after this is merged,
Use this image in the graphql package.

PR #6978, opened 2022-01-31
graphql - add requests-toolbelt dependency

Itay4:
## Status
Ready

Related PR: https://github.com/demisto/content/pull/16358

## Description
was added in https://github.com/graphql-python/gql/pull/244


wissamg (approved):
Guys,

 We do not move the python version forward already if we are updating!?

PR #7314, opened 2022-03-27
Bumping up chrome to resolve CVE 2022 1096

wissamg:
## Status
Ready

## Related Issues
Related: [47611](https://github.com/demisto/etc/issues/47611)

## Description
bumping up chrome version to cover over High CVE-2022-1096


glicht (changes requested):
We should merge this after we update the python PR and then use the new base image.

glicht:
> We should merge this after we update the python PR and then use the new base image.

@wissamg see my comment above

PR #7686, opened 2022-05-11
add install command for kubectl and upgrade the image

wissamg:
## Status
Ready

## Description
installing kubectl command and upgrade the image using apt-get upgrade -y


wissamg:
> Don't you want to remove the installations?

You are I will do that

PR #7826, opened 2022-05-31
add cowsay to gitlab content build image

yaakovi:
## Status
Ready

## Description
for better UX in build failures.


avidan-H (approved):
I've never approved of anything more than I do this change. If I could approve twice I would.

PR #7919, opened 2022-06-12
xsoar-tools sdk v1.5.4 hotfix

dorschw:
https://jira-hq.paloaltonetworks.local/browse/CIAC-3277

PR #7948, opened 2022-06-15
fix warning bug

dorschw:
We used to do `pipenv lock -r > requirements.txt`, which sometimes wrote warnings into the file, then installing the file [failed](https://app.circleci.com/pipelines/github/demisto/dockerfiles/24688/workflows/5dbe8e50-56b2-492e-a3f7-370c49532b54/jobs/30677) as the warnings obviously do not match the expected syntax of `requirements.txt`.

### DIFF
--- a/docker/build_docker.sh
+++ b/docker/build_docker.sh
@@ -159,7 +159,7 @@ function docker_build {
             return 1
         fi
         pipenv --rm || echo "Proceeding. It is ok that no virtualenv is available to remove"
-        PIPENV_YES=yes pipenv lock -r --no-header> requirements.txt
+        PIPENV_YES=yes pipenv run pip freeze > requirements.txt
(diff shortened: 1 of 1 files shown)


DeanArbel (approved):

glicht:
Can you add the actual docker image, so we see that it fixes the issue.

dorschw:
that was the thing that fixed this PR
<https://github.com/demisto/dockerfiles/pull/7919>, in this commit
<https://github.com/demisto/dockerfiles/blob/f5826c2bf9d00fec09eb7eb53f6ab41cd06c0f99/docker/build_docker.sh#L162>
(I
have since then reverted the change to be implemented in the PR we're
discussing on, and not on the xsoar-tools test PR)

glicht (approved):

wissamg (approved):

PR #8214, opened 2022-07-20
Armorblox

SamhithaTatipalli:
## Status

## Related Issues
Related: link to the issue

## Description


SamhithaTatipalli:
Hi Team,

Could you please let us know the ETA for this PR.

Thanks and Regards,
Samhitha.

ShahafBenYakir:
@SamhithaTatipalli As the template suggests we need a matching PR in the [content](https://github.com/demisto/content) repo, do you have such? if so please comment on it with this PR link, if not, we will review this PR when you'll have a matching PR in the content repo.

SamhithaTatipalli:
@ShahafBenYakir, Yes, we have a PR to link with it but it is basically in review internally- Once it is passed internally - I would link it post raising a PR to demisto/content.

SamhithaTatipalli:
@ShahafBenYakir, Please find the matching PR in the content repo: https://github.com/armorblox/content/pull/61.

ShahafBenYakir:
@SamhithaTatipalli We review docker PRs together with their corresponding content PRs so the same reviewer will review both. In your case, I can see you made some changes on your internal repo but you have not yet created a pull request to the [demisto/content repo](https://github.com/demisto/content). Once you'll have that content PR open, tag this PR in the content PR.

SamhithaTatipalli:
@ShahafBenYakir, yes we have raised a PR, but the PR Unit testing is failing since image is not present in the docker hub. We are using that image in our content repo. Unless and until if this PR gets approved, unit test will always fail. Please find the link for reference:https://github.com/demisto/content/pull/20455. 

Could you please assist me on this?

docker/armorblox/Pipfile
```diff
@@ -0,0 +1,12 @@
+[[source]]
+url = "https://pypi.org/simple"
+verify_ssl = true
+name = "pypi"
+
+[packages]
+armorblox-sdk = "==0.1.2"
```
ShacharKidor, line 7: Is there a reason for setting the version to 0.1.2? 
Locking the version to a specific version number will prevent an automatic update when the package is updated to a more advanced version. someone would need to update the version manually. 

Another option is to set it as follows:
`armorblox-sdk = "*"`
This way the version will be updated automatically. 
@wissamg - what do you think will be better here?

docker/armorblox/Pipfile
```diff
@@ -0,0 +1,12 @@
...
+
+[dev-packages]
+
+[requires]
+python_version = "3.9"
```
ShacharKidor, line 12:
```suggestion
python_version = "3.10"
```

ShacharKidor (changes requested):
Hi @SamhithaTatipalli, 
Good Job!
I added two comments please review them. 

@wissamg - Please review it as well, I would like your approval before merging that. 

Thank you!

docker/armorblox/Pipfile
urajat: @ShacharKidor While we try to maintain backward compatibility for our Python SDK, we do not guarantee it at this point as the SDK is still maturing. That's why we wanted to assure compatibility with a specific version rather than allowing arbitrary upgrades which can end up breaking functionality for existing users. Eventually, we want to move to the wildcard option at least for patch version changes.

ShacharKidor: ok so let's keep it this way :)

docker/armorblox/Dockerfile
```diff
@@ -0,0 +1,8 @@
+
+FROM demisto/python3:3.10.4.30607
```
wissamg, line 2:
```suggestion
FROM demisto/python3:3.10.5.31928
```

ShacharKidor:
I opened a new PR with fixes of the pipfile.lock and solved conflicts [8431](https://github.com/demisto/dockerfiles/pull/8431). 
Closing the PR.

PR #8412, opened 2022-08-22
Updated the python version that the images uses (to 3.9)

anas-yousef:


docker/dxl/Pipfile
```diff
@@ -12,4 +12,4 @@ dxltieclient = "*"
 
 
 [requires]
-python_version = "3.7"
\ No newline at end of file
+python_version = "3.10"
```
dorschw, line 15: 3.9? If you change it, please re-`lock`

dorschw (changes requested):

PR #8460, opened 2022-08-29
Access data v2

sriram7797:
## Status
Ready

## Related Content Pull Request
Related PR: https://github.com/demisto/content/pull/20823

## Description
Docker image `demisto/accessdata:1.1.0.32812` needs to rebuild as dependent module is changed(cached as part of docker image)

The image was originally added in https://github.com/demisto/dockerfiles/pull/8326/files.

Since then, changes were made in branch `AccessDataV2`. We need to incorporate these changes into a latter tag of the `demisto/accessdata`.

PR #8473, opened 2022-08-30
Access data v2

sriram7797:
## Status
Ready

Related PR: https://github.com/demisto/content/pull/20823

## Description
requirements.txt updated with latest dependent commit version


sriram7797:
@kgal-pan Please review and give production docker image to update in content PR (https://github.com/demisto/content/pull/20823)

docker/accessdata/requirements.txt
```diff
@@ -0,0 +1 @@
+git+https://github.com/AccessDataOps/FTK-API-SDK.git@a3d9281586f25fdbef9cfab7e0f0082321b8afa6
```
YuvHayun, line 1: Why do you want to use this particular commit rather than the branch name?

sriram7797: previously i had branch name only, but when changes were made in FTK-API-SDK, docker build is getting failed because there would be no change in dockerfiles/docker/accessdata folder. So changed to commit hash from branch name. (once docker image is built, sdk will be cached in the image as a layer). for each change in the ftk-api-sdk, i will update the corresponding commit hash value in the requirements.txt file, so that build will not fail and will get new image tag for each.(adding @kgal-pan for reference, we had discussion)

kgal-pan: @YuvHayun  - For reference, see [the PR](https://github.com/demisto/dockerfiles/pull/8460/files) where the branch name was used. As you can see, the branch name was used there but since the branch was updated, there is no diff in the `Dockerfile`. Is there a way to update the Docker image when there are no diffs in the `Dockerfile`? 

I believe this is the reason @sriram7797 used a commit hash instead of branch name.

sriram7797: @kgal-pan Yes that's the reason i used commit. 

basically each command in the Dockerfile will be cached as the layer in the machine on which image is built. when we build docker image again without changes in the Dockerfile, it will take the cached layer from machine (command wont be executed to build the docker image). we need to build docker image with **--no-cache** option. i am just not sure this could also be problem. (came across this scenario when building test build locally using ./build_docker.sh)

YuvHayun (approved):

YuvHayun:
@sriram7797
I approve this PR and I'll merge it when the other one is merged.

PR #9198, opened 2022-10-18
update pymisp

dansterenson:
pymisp is using the extract-msg dependency which requires a GPL license. Since we do not have this license, we decided to manually remove the extract-msg dependency and support only .eml files (and not .msg).

## Related Issues
Related: [link to the issue](https://jira-hq.paloaltonetworks.local/browse/XSUP-17362)


docker/pymisp2/Dockerfile
```diff
@@ -5,4 +5,6 @@ COPY requirements.txt .
 
 RUN pip install --upgrade pip \
  && apk --update add --no-cache --virtual .build-dependencies python3-dev gcc linux-headers libc-dev libffi-dev\
- build-base wget git && pip install --no-cache-dir -r requirements.txt && apk del .build-dependencies
+ build-base wget git && pip install --no-cache-dir -r requirements.txt && apk del .build-dependencies && pip uninstall -y extract-msg\
+ && pip uninstall -y pcodedmp\
+ && find . -type f -name 'emailobject.py' -exec sed -i -e '/from extract_msg import openMsg/d; /from extract_msg.message import Message as MsgObj/d; s/msg_obj: MsgObj/msg_obj/g' {} \;
```
glicht, line 10: Using `find ... exec` is not a good choice. As if it doesn't find the file it doesn't fail. You actually don't need `find`. You can simply use a glob. Use the following:

 ```suggestion
 && sed -i -e '/from extract_msg import openMsg/d; /from extract_msg.message import Message as MsgObj/d; s/msg_obj: MsgObj/msg_obj/g' /usr/local/lib/python3*/site-packages/pymisp/tools/emailobject.py
```

This will also fail if no file is found (for example if pymisp change emailobject.py file's name).

glicht (changes requested):
Please add verify.py file to do a basic check that it is possible to load pymisp

docker/pymisp2/Dockerfile
dansterenson: Done

glicht (approved):

PR #9470, opened 2022-11-15
add python3-ubi image

GuyAfik:
## Status
Ready

## Related Issues
Related: link to the issue

## Description
add python3-rhel image which will be a base image for python3 that is based on redhat linux


docker/python3-rhel/Dockerfile
```diff
@@ -0,0 +1,85 @@
+# python3.10.8 at the moment - if updated please update this comment
+ARG BASE_UBI_TAG=8.6
+
+FROM registry.access.redhat.com/ubi8:${BASE_UBI_TAG} as build
```
JudahSchwartz, line 4: would put ubi8 as var also

docker/python3-ubi/verify.py
```diff
@@ -0,0 +1,15 @@
+import warnings
```
glicht, line 1: This file should be a symbolic link to the python3/verify.py file as we do in python3-deb. See for example: https://github.com/demisto/dockerfiles/blob/master/docker/python3-deb/verify.py

docker/python3-ubi/Pipfile
```diff
@@ -0,0 +1,21 @@
...
+
+[packages]
+requests = {extras = ["socks"],version = "*"}
+olefile = "*"
+python-dateutil = "*"
+dateparser = "*"
+pyyaml = "*"
+funcy = "*"
+tldextract = "*"
+pip = "*"
```
glicht, line 16: Why is this file different from the Pipfile used by python3? 

See: https://github.com/demisto/dockerfiles/blob/master/docker/python3/Pipfile

glicht (changes requested):

docker/python3-ubi/Dockerfile
```diff
@@ -0,0 +1,84 @@
+# python3.10.8 at the moment - if updated please update this comment
+ARG BASE_UBI_TAG=8.6
```
glicht, line 2: Why not use 8.7?

GuyAfik: updated to 8.7

docker/python3-ubi/verify.py
GuyAfik: done

docker/python3-ubi/Dockerfile
```diff
@@ -0,0 +1,84 @@
...
+    echo '/usr/local/lib' > /etc/ld.so.conf && \
+    ldconfig
+
+COPY requirements.txt .
+RUN pip3.10 install -r requirements.txt
+
+FROM registry.access.redhat.com/${BASE_UBI_VERSION}:${BASE_UBI_TAG}
+
+RUN dnf update -y --nodocs && \
+    dnf clean all && \
```
glicht, line 51: Let's also add procps-ng so we have `ps` available for troubleshooting. You can add:
```
dnf install --nodocs -y  procps-ng
```

GuyAfik: done

docker/python3-ubi/Pipfile
GuyAfik: done

glicht (approved):
Looks good

PR #10665, opened 2022-12-26
Edit openssl cnf

thefrieddan1:
to handle UnsafeLegacyRenegotiation

## Status

## Related Issues
Related: https://jira-hq.paloaltonetworks.local/browse/CIAC-4962

## Description
As a result of SSL Legacy renegotiation error, Adding UnsafeLegacyRenegotiation option to the SSL conf file.


docker/python3/Dockerfile
```diff
@@ -14,3 +14,16 @@ RUN apk --update add --no-cache --virtual .build-dependencies python3-dev build-
 
 RUN addgroup -g 4000 demisto \
   && adduser -u 4000 -G demisto -D demisto -s /bin/sh 
+
+# Handling the issue described here https://knowledgebase.paloaltonetworks.com/KCSArticleDetail?id=kA10g000000POJ0CAO&lang=en_US%E2%80%A9
+# by enabling UnsafeLegacyRenegotiation
+RUN echo 'openssl_conf = openssl_init\
```
JudahSchwartz, line 20: can we have this in the folder as a file and COPY it in?

JudahSchwartz:
reminder to add this to python3-deb and python3-native also

docker/python3/Dockerfile
GuyAfik: +1, there are other docker-images which should be updated, better to do in an external file and copy it

GuyAfik:
> reminder to add this to python3-deb and python3-native powershel also

this should be done on **python3-ubi** and not on py3-native, py3-native "inherits" from python3-ubi

PR #11105, opened 2023-01-06
update fingerprint

glicht:
## Status
Ready

## Related Issues
Related: link to the issue

## Description
* Updating the trust key.
* Updating unzip to debian as unrar is not available anymore in alpine. Also see that build is working good.


JudahSchwartz:
@GuyAfik how were we able to support unzip in the native docker image? Do we need to make changes here?

GuyAfik:
> @GuyAfik how were we able to support unzip in the native docker image? Do we need to make changes here?

we were able to install it from fedora-epel repo, here: https://github.com/demisto/dockerfiles/blob/master/docker/py3-native/Dockerfile#L68

regardless the script that uses this image is ignored at the moment. https://github.com/demisto/content/blob/master/Tests/docker_native_image_config.json#L37

There isn't anything needs to be done in the py3-native as long as no new OS dependency was added

JudahSchwartz (approved):

PR #11572, opened 2023-01-18
Py3ews - Activated the legacy section in the openssl.cnf

ShacharKidor:
## Status
Ready (Reason for hold)

## Related Content Pull Request
Related PR: link to the PR at demisto/content

## Related Issues
Two bugs:
Related: [XSUP-20359](https://jira-hq.paloaltonetworks.local/browse/XSUP-20359),
Related: [XSUP-20247](https://jira-hq.paloaltonetworks.local/browse/XSUP-20247).

An old bug that I think is also related: 
[XSUP-19122](https://jira-hq.paloaltonetworks.local/browse/XSUP-19122).

## Description
In the above mention bugs, EWS Mail Sender integration failed after upgrading the docker image to: demisto/py3ews:1.0.0.43044. 
The error in the logs is: 
```
  File "/usr/local/lib/python3.10/hashlib.py", line 123, in __get_builtin_constructor
    raise ValueError('unsupported hash type ' + name)
ValueError: unsupported hash type md4
```
It appears that some hashing algorithms with MD4 among them were disabled as obsoleted, which caused that error. 
To solve that we activated the legacy section in the openssl.cnf file (re-enabled md4 to hashlib) - based on the solution described here: https://github.com/ecederstrand/exchangelib/issues/608.

PR #11574, opened 2023-01-19
py-ews - Added check that md4 is enabled to verify.py

ShacharKidor:
## Status
Ready

Related PR: [11572](https://github.com/demisto/dockerfiles/pull/11572).

## Description
As we did in py3ews, I added a check to the verify.py of the py-ews image to make sure that the MD4 is enabled. 
That way we should catch problems in future upgrades.


ShacharKidor:
Closing the PR. 
Since md4 isn't in the list of hashlib.algorithms_available this shouldn't be an issue.

PR #11591, opened 2023-01-22
add more_itertools

dorschw:


GuyAfik (commented):
Looks great!

Please add imports of `more-itertools` to the `verify.py` of the python3 image so we can actually make sure its successfully being imported in the container. 

https://github.com/demisto/dockerfiles/blob/master/docker/python3/verify.py

GuyAfik (approved):

PR #12822, opened 2023-02-23
Updated poetry.lockfor devo image

Shellyber:
## Status
Ready

Related PR: https://github.com/demisto/content/pull/24842

## Description
Updated the packages because of the following issue: [AttributeError: module 'lib' has no attribute 'OpenSSL_add_all_algorithms'](https://stackoverflow.com/questions/74981558/error-updating-python3-pip-attributeerror-module-lib-has-no-attribute-openss)


ShahafBenYakir:
I am confused, why did we added cryptography to this image as a fix?

Shellyber:
We didn't, It already existed in the docker image.
But the latest update that happened upgraded the version, which broke the integration
I just reverted back to the previous version

JudahSchwartz:
@Shellyber if we need the version locked we also need to specify it in the dependabot config, otherwise dependabot will upgrade it anyway

JudahSchwartz:
@Shellyber we cant make the fix on the integration side instead?

Shellyber:
@JudahSchwartz No, this is a partner integration so I can't really fix it locally
Also when trying to update all (so we won't need to lock it) it caused a different problems with their SDK
So the best thing is to lock the version

PR #12934, opened 2023-03-01
add script

JudahSchwartz:
## Status
Ready 

## Description
This is a script that finds dockerimages that should be deprecated but arent. Meant to be run locally


ShahafBenYakir (approved):
Let's add a description comment at the beginning of the file on how/where to run this and that it should not be deleted.

PR #14899, opened 2023-04-19
dynamic versioning

JudahSchwartz:
We have several docker images that would be convenient if we could update their versions dynamically, instead of just adding manually to build.conf, like python images, powershell images, demisto-sdk (and others that are meant to track a specific python package)

This is generic ingrastructure to support a `dynamic_version.sh` file, which when provided will be run on the image. Any output of that script will be used as the version

This pr also removes the special code i wrote in the auto_update flow to update the python versions, as it isnt neccessary anymore

solves https://jira-hq.paloaltonetworks.local/browse/CIAC-3339


docker/build_docker.sh
```diff
@@ -207,7 +207,23 @@ function docker_build {
         --label "org.opencontainers.image.authors=Demisto <containers@demisto.com>" \
         --label "org.opencontainers.image.version=${VERSION}" \
         --label "org.opencontainers.image.revision=${CIRCLE_SHA1}"
+
+    if [[ -e "dynamic_version.sh" ]]; then
+      echo "dynamic_version.sh file was found"
+      dynamic_version=$(docker run --rm -i "$image_full_name" sh < dynamic_version.sh)
+      echo "dynamic_version $dynamic_version"
```
tkatzir, line 214: Let's log all the relevant variables
```suggestion
      echo "image_full_name: ${DOCKER_ORG}/${image_name}:$dynamic_version.${REVISION}"
```

docker/build_docker.sh
```diff
@@ -207,7 +207,23 @@ function docker_build {
...
+      VERSION="${dynamic_version}.${REVISION}"
+      image_full_name="${DOCKER_ORG}/${image_name}:${VERSION}"
+
+      # add the last layer and rebuild. Everything shuld be cached besides this layer
+      echo "ENV DOCKER_IMAGE=$image_full_name" >> "$tmp_dir/Dockerfile"
+      docker build -f "$tmp_dir/Dockerfile" . -t ${image_full_name} \
+        --label "org.opencontainers.image.authors=Demisto <containers@demisto.com>" \
+        --label "org.opencontainers.image.version=${VERSION}" \
+        --label "org.opencontainers.image.revision=${CIRCLE_SHA1}"
```
tkatzir, line 223: Is the revision `${CIRCLE_SHA1}` or `${REVISION}`?

tkatzir (approved):
Approved, with a comment ad a question. :-)

docker/build_docker.sh
JudahSchwartz: ${CIRCLE_SHA1} contains the SHA1 hash of the commit that triggered the current build in CircleCI. 
${REVISION} is the last 5 digits we have at the end of VERSION
im not sure why we call this label by this name, but this is how it was before also

docker/build_docker.sh
JudahSchwartz: will add it the next line

PR #15734, opened 2023-05-07
Add hypothesis to native

kgal-pan:
## Status
Ready

Related PR: https://github.com/demisto/content/pull/24824

## Description
Add `hypothesis` package used for unit testing. This package was available previously but was removed in a recent upgrade and broke https://github.com/demisto/content/actions/runs/4907742444/jobs/8763044652?pr=24824.


GuyAfik (approved):
@kgal-pan looks great to me, please wait @JudahSchwartz 's cr before merging.

JudahSchwartz:
why do we want to add this package? Which docker image are we supporting with this?
testing dependencies should not be in production images, they should be added during demisto-sdk lint by the demisto-sdk

kgal-pan:
@JudahSchwartz - Not sure which dependency was updated in the last month but `hypothesis` was removed and is causing failures in [this job](https://github.com/demisto/content/actions/runs/4908794700/jobs/8764786616?pr=24824#step:8:228)  now (the job was running fine last month). We can see that this package was there previously because it shows up in the `poetry.lock` on `master`:

```bash

> grep "hypothesis" -rnw ~xsoar_dockerfiles

/Users/kgal/dev/demisto/dockerfiles/docker/py3-native/poetry.lock:1839:test = ["hypothesis (>=5.5.3)", "pytest (>=6.0)", "pytest-xdist (>=1.31)"]
/Users/kgal/dev/demisto/dockerfiles/docker/py3-native/poetry.lock:1959:test = ["attrs (>=20.2.0)", "coverage[toml]", "hypothesis (>=6.36)", "psutil (>=5.9)", "pybind11", "pytest (>=6.2.5)", "pytest-cov (>=3.0.0)", "pytest-timeout (>=2.1.0)", "pytest-xdist (>=2.5.0)", "python-dateutil (>=2.8.1)", "python-xmp-toolkit (>=2.0.1)", "tomli"]
/Users/kgal/dev/demisto/dockerfiles/docker/py3-native/poetry.lock:2403:testing = ["argcomplete", "hypothesis (>=3.56)", "mock", "nose", "pygments (>=2.7.2)", "requests", "xmlschema"]
``` 

Please advise.

JudahSchwartz (changes requested):
it could be at some point some package made a mistake and added this under dependencies instead of test dependencies, so this package moved into the native image. 
We only install production dependencies in the docker
Currently we only support [these test dependencies](https://github.com/demisto/demisto-sdk/blob/master/demisto_sdk/commands/lint/resources/pipfile_python3/Pipfile)
If we want to support a test package it needs to be added there.
Also i saw in your pr native image is defined in the dockerimage tag and should be switched.

JudahSchwartz:
actually i see hypothesis is in that list... @ilaner @dorschw do we know why precommit fails? Are we installing that pipfile there?

PR #16429, opened 2023-05-22
Lock exchangelib to version 4.8.0

amshamah419:
## Status
Ready

## Related Issues
Related: https://jira-hq.paloaltonetworks.local/browse/XSUP-24410

## Description
Echangelib 5.0.0 introduced changes to xml handling in autodiscovery. This version is the latest before that change.

### DIFF
--- a/docker/py3ews/Pipfile
+++ b/docker/py3ews/Pipfile
@@ -6,7 +6,7 @@ verify_ssl = true
 [dev-packages]
 
 [packages]
-exchangelib = "*"
+exchangelib = "4.8.0"
 chardet = "*"
 dateparser = "*"
 
--- a/docker/py3ews/Pipfile.lock
+++ b/docker/py3ews/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "a4cbf07b48679592d04f17739c2e078b27b556d7ffd86c04dc321c521a321e8f"
+            "sha256": "6ce4a0026be7c89a271805ebe8ca2f786bbfcdb707356d33d0897f863ee53c3f"
         },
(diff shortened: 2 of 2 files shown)


docker/py3ews/Pipfile
```diff
@@ -6,7 +6,7 @@ verify_ssl = true
 [dev-packages]
 
 [packages]
-exchangelib = "*"
+exchangelib = "4.8.0"
```
JudahSchwartz, line 9: can you comment here why this is locked? Should we be locking this in the image we're working on for ewsv2? FYI @ostolero

amshamah419: I'm not sure we need to. Basically we have a customer reporting that in 1.2.1 of EWS the integration fails to fetch. It appears related to how attachments are processed and in exchangelib 5.0.0, there was a sizable change to how the XML file which holds these attachments are parsed. I'm trying to freeze the version because I believe the changes that were made is effecting how we handle the file.

JudahSchwartz: if we do end up needing to lock lets leave a comment in the file so we wont be wondering in a year

amshamah419: That hardly sounds fun. Where's your spirit of adventure?
```suggestion
exchangelib = "4.8.0" # Behind whispers of codes, a secret's veil unraveled; Exchangelib's version frozen, a riddle unsolved.
```

ostolero: I'll just add from my experience advancing the EWS V2 to the new exchange-lib.
I think these changes are better handled by fixing the code, freezing prevents the BC issues but also prevents new advances our clients will want in the future... I see EWSMailSender also uses this docker so take that into consideration.

PR #19081, opened 2023-08-13
update py3-native performence

GuyAfik:
## Status
Ready

## Description
Update py3-native performence


tkatzir (approved):

GuyAfik:
need to build this image urgently, force merging because it requires +1 of the code owner that is not here ATM.

PR #19210, opened 2023-08-15
python3-ubi - add python version env var

GuyAfik:
## Status
Ready

## Description
Add the `PYTHON_VERSION` env var to the python3-ubi image.

### DIFF
--- a/docker/python3-ubi/Dockerfile
+++ b/docker/python3-ubi/Dockerfile
@@ -100,7 +100,7 @@ Options = UnsafeLegacyRenegotiation\n' > /tmp/ssl.cnf \
 RUN groupadd -g 4000 demisto && \
     useradd -u 4000 -g demisto demisto -s /bin/sh
 
-ENV PYTHON_VERSION=${PYTHON_VER}
+ENV PYTHON_VERSION="3.10.8"
 
 CMD ["python3"]
 
--- a/docker/python3-ubi/Dockerfile
+++ b/docker/python3-ubi/Dockerfile
@@ -1,5 +1,6 @@
 ARG BASE_UBI_TAG=9.2
 ARG BASE_UBI_VERSION=ubi9
+ARG PYTHON_VER = "3.10.8"
 
 FROM registry.access.redhat.com/${BASE_UBI_VERSION}:${BASE_UBI_TAG} as build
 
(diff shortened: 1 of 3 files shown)


docker/python3-ubi/Dockerfile
```diff
@@ -98,6 +98,8 @@ Options = UnsafeLegacyRenegotiation\n' > /tmp/ssl.cnf \
 RUN groupadd -g 4000 demisto && \
     useradd -u 4000 -g demisto demisto -s /bin/sh
 
+ENV PYTHON_VERSION=$(python3 -c 'import platform; print(platform.python_version())')
```
JudahSchwartz, line 101: does this work? thought it cant be dynamic.
I think we should do ARG PYTHON_VERDION=3.10.8 that we use for this variable and also for the download and install iteslf. Or maybe use ENV if the env cant depend on the arg?

GuyAfik: you are right it can't be dynamic

docker/python3-ubi/Dockerfile
```diff
@@ -19,11 +19,13 @@ RUN dnf upgrade -y --nodocs && \
     dnf clean all && \
     rm -rf /var/cache/dnf
 
-# install python3
-RUN wget https://www.python.org/ftp/python/3.10.8/Python-3.10.8.tgz && \
+ARG PYTHON_VER=3.10.8
```
JudahSchwartz, line 22: any way to declare as an ENV here and use it on line 27 so we dont need to declare 3.10.8 twice? 
Id also use a symlink to the dynamic_version script we have in python3 so we dont need to repeat 3.10.8 in the build conf either

JudahSchwartz (approved):
Approved except my last comment. See if thats doable, not sure it is

JudahSchwartz:
also I would add to changelog that you added this env var

GuyAfik:
> also I would add to changelog that you added this env var

the change log is only for py3-native

docker/python3-ubi/Dockerfile
GuyAfik: we use multistage, is it possible? regarding dynamic_version this can be done

GuyAfik: added dynamic version symlink

PR #19390, opened 2023-08-23
update sdk version to 1.19.1

adi88d:
## Status
Ready

## Related Issues
Related: link to the issue

## Description


docker/xsoar-tools/Pipfile
```diff
@@ -5,7 +5,7 @@ name = "pypi"
 
 [packages]
 demisto-py = "*"
-demisto-sdk = "==1.17.2" # this package should be updated manually. No reason this is locked
+demisto-sdk = "*"
```
JudahSchwartz, line 8: didnt we want this locked ? @dorschw

dorschw: indeed, it should be locked.

GuyAfik: why is it locked? why can't we update it with depandabot?

dorschw: we can, but want to control it manually.

JudahSchwartz: hence the comment

GuyAfik: @adi88d i still dont understand why we want to control it manually but go ahead and lock it.
i don't see any reason why we need to update it manually each time, it makes some of our content scripts to run with old SDK versions which contain bugs that were fixed in later versions

dorschw: See [here](https://panw-global.slack.com/archives/C03L972AER1/p1692798368468139)

dorschw, line 8: change to `1.19.1`

dorschw (changes requested):

docker/xsoar-tools/Pipfile
```diff
@@ -5,7 +5,7 @@ name = "pypi"
 
 [packages]
 demisto-py = "*"
-demisto-sdk = "==1.17.2" # this package should be updated manually. No reason this is locked
+demisto-sdk = "1.19.1" # this should only be modified manually. only specify a version, do not use *.
```
dorschw, line 8:
```suggestion
demisto-sdk = "==1.19.1" # this should only be modified manually. only specify a version, do not use *.
```

dorschw (approved):

dorschw (approved):

PR #20515, opened 2023-09-26
Add packages for mypy_boto3

mmhw:
## Status
Ready

Related PR: [link to the PR at demisto/content](https://github.com/demisto/content/pull/29819)

## Related Issues
Related: [link to the issue](https://jira-hq.paloaltonetworks.local/browse/CIAC-8236)

## Description
To avoid the `MyPy` errors that `"boto3.client" is not valid as a type`, it is necessary to use `mypy_boto3` for the types.

### DIFF
--- a/docker/boto3py3/Pipfile
+++ b/docker/boto3py3/Pipfile
@@ -7,6 +7,9 @@ verify_ssl = true
 
 [packages]
 boto3 = "*"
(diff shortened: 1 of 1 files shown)


JudahSchwartz (changes requested):
mypy packags shouldnt be in our production environment

PR #21018, opened 2023-10-17
Fix gdetect pip version

glimps-jbo:
## Status
Ready

## Description
Just a quick fix in our pip requirement to freeze the library version (before some important changes in the library)

Thank you.


JudahSchwartz:
you need to also lock the version in dependabot config here .github/dependabot.yml or dependabot will bump it

omerKarkKatz:
@glimps-jbo Why do you want to lock it on the older version? 
Is the new version presenting breaking changes?

glimps-jbo:
> @glimps-jbo Why do you want to lock it on the older version? Is the new version presenting breaking changes?
The new version will change some calls and exceptions so we want to prevent unwanted upgrade. 
After that, we will update our demisto code to use the new version.



> you need to also lock the version in dependabot config here .github/dependabot.yml or dependabot will bump it

We did not add any entry in dependabot.yml, does this mean we won't have any automatic bump ? (It's the behaviour we want until we align the code)

omerKarkKatz:
@glimps-jbo I see that there is an entry for your docker.
image

JudahSchwartz:
if you used our script that automatically adds the image then we append to that file. I think removing it from that file isnt enough, because depenadbot has some strange behavior. I think we should leave it there and add an ignore

omerKarkKatz:
Hi @glimps-jbo, we haven’t heard from you in a while.
Do you need any help with the pull request?

Please feel free to reach out to me here or on Slack.
Thanks again for contributing to our repo, hope to hear from you soon.

omerKarkKatz:
Hi @glimps-jbo,
Thank you for your contribution. 
I have not received a response from you in the last two weeks, so I will close your PR now.
Please feel free to re-open it when you are available to continue. 
Thanks again.

PR #23236, opened 2024-01-07
var Files

tkatzir:
## Status
Ready

Related PR: https://github.com/demisto/content/pull/30849


JudahSchwartz (commented):
this script will now be on content side? 
why not let content have control over these files?

tkatzir:
> this script will now be on content side? why not let content have control over these files?

Yes. we'll run the Chrome from Python.

When the Docker container is hardened, we run with the `demisto` user (and not `root`,) so we can't create these files.

JudahSchwartz:
can we move these files under the demisto directory? (should be cwd). This is the dir that we use to create files for file_results and other files we create

tkatzir:
> can we move these files under the demisto directory? (should be cwd). This is the dir that we use to create files for file_results and other files we create

We need these files to be retained between runs of the Looper, so no.

JudahSchwartz (approved):
lets double check that this is needed
https://github.com/demisto/content/pull/19512 I remember this bug that was caused because files dont get cleaned up between runs

PR #26351, opened 2024-05-01
add pydantic to `py3-tools`

dorschw:
related: https://jira-dc.paloaltonetworks.com/browse/CIAC-10398
- `py3-tools`: added pydantic `>=2.7.1` to prevent unintended downgrading to v1.
- `py3-tools-ubi`: added pydantic `>=2.7.1` to prevent unintended downgrading to v1.
- `auth-utils`: locked (existing dependency) pydantic to `>=2.7.1` to prevent unintended downgrading to v1.
- `fastapi`: locked to use Pydantic V1 as `fastapi_utils` (another dependency) [doesn't yet work with Pydantic v2](https://github.com/dmontagu/fastapi-utils/issues/276). 
- `py3-native`: changed pydantic to be locked to use `>=2.7.1` instead of `>=2.0.0`, to ensure compatibility with other images.


docker/auth-utils/pyproject.toml
```diff
@@ -7,7 +7,7 @@ readme = "README.md"
 
 [tool.poetry.dependencies]
 python = "~3.10"
-pydantic = "*"
+pydantic = "^2.7.1"
```
JudahSchwartz, line 10: can we add as >= not ^ ?  want to allow the bump to 3

docker/fastapi/Pipfile
```diff
@@ -10,6 +10,7 @@ fastapi = "*"
 uvicorn = "*"
 fastapi_utils = "*"
 httpx = "*"
+pydantic = ">=1.0.0,<2.0.0" # prevent unintended upgrade to v2
```
JudahSchwartz, line 13: you also need to lock this in the dependabot config otherwise dependabot will bump this i believe

docker/py3-native/pyproject.toml
```diff
@@ -73,7 +73,7 @@ netaddr = "*"
 python-libnmap = "*"
 websockets = "*"
 dnspython = "*"
-pydantic = ">=2.0.0"
+pydantic = "^2.7.1"
```
JudahSchwartz, line 76: would also prefer >= 2.7.1

JudahSchwartz (approved):

dorschw:
@yuvalbenshalom please force-merge. Reason is [CIAC-10470](https://jira-dc.paloaltonetworks.com/browse/CIAC-10470).

PR #27435, opened 2024-06-04
bump websockets

JudahSchwartz:
bump websockets > 12

```
>>> from websockets.sync.client import connect
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'websockets.sync
```


shmuel44 (approved):
Nice!

PR #28306, opened 2024-06-24
Native Image: Match dependencies, avoid discrepancies

dorschw:
fixes: [CIAC-10937](https://jira-dc.paloaltonetworks.com/browse/CIAC-10937)

Matching native image and contained images:

1. Locked versions where we were already using newer versions (e.g. was `*` but in lock file `>1.2.3`, changing to `>=1.2.3` or `>=1.2.0` won't matter, and make it safer with native image)
- crypto: locked `cryptography = ">=39.0.1"` 
- google-api-py3: locked `cryptography = ">=39.0.1"`
- netutils: locked `websockets = ">=12.0.0"`

2. Native image: Same locking as we're using >= the now-locked versions
- Bumped urllib3 to 2.2.2 
- Bumped future to >=1.0.0 
- Locked requests-ntlm = "==1.1.0"
- Locked slackclient = ">2.0"
- Locked aiodns = ">1.0"
- Bumped pydantic = ">=2.7.1"

3. Native image: bumped exchengelib to >=5.4.1 (bugfix)
4. Native image: added a few more dependencies from contained images


dorschw:
@yuvalbenshalom please force-merge, the build failures are (1) not caused by content items I added, (2) irrelevant in a content context, (3) the first one is unfixable and should be ignored.

PR #29728, opened 2024-07-16
GitHub Action to Compare Dependencies 

shmuel44:
## Related Issues
Related: [link](https://jira-dc.paloaltonetworks.com/browse/CIAC-11190) to the issue

## Description
Added a GitHub action to check if there is a difference between dependencies of the native Docker image and all the other images that are supported by the native image.

image


.github/workflows/sync_native_dependency.yml
```diff
@@ -0,0 +1,27 @@
...
+        with:
+          python-version: "3.11"
+      - name: Install dependencies with pipenv
+        run: |
+          python -m pip install --upgrade pip
+          pip install pipenv
+          pipenv install
+      - name: Run compare_dependency_constraints Script.
+        run: |
+          ls
```
dorschw, line 23: is this needed? If it is, would be clearer if we moved it to a different stage

.github/workflows/sync_native_dependency.yml
```diff
@@ -0,0 +1,27 @@
...
+          echo "==== $(date): Starting environment setup... ===="
+          cd "$GITHUB_WORKSPACE"
```
dorschw, line 25: needed? 🤔

.github/workflows/sync_native_dependency.yml
```diff
@@ -0,0 +1,27 @@
...
+          python-version: "3.11"
+      - name: Install dependencies with pipenv
+        run: |
+          python -m pip install --upgrade pip
+          pip install pipenv
+          pipenv install
+      - name: Run compare_dependency_constraints Script.
+        run: |
+          ls
+          echo "==== $(date): Starting environment setup... ===="
```
dorschw, line 24:
```suggestion
```
we don't really setup much here, maybe in the `Install dependencies` step this would make more sense, but removing it would make just as much sense

.github/workflows/sync_native_dependency.yml
```diff
@@ -0,0 +1,27 @@
...
+          cd "$GITHUB_WORKSPACE"
+          echo "==== $(date): Running chceking Native dependency ===="
```
dorschw, line 26:
```suggestion
```
since this is (now) the only thing in the file :)

utils/compare_dependency_constraints.py
```diff
@@ -49,8 +50,25 @@ class Discrepancy(NamedTuple):
     in_native: str | None = None
 
     def __str__(self) -> str:
-        return f"{self.dependency}: {self.in_image or 'missing'} in {self.image}, {self.in_native or 'missing'} in native"
+        return f"{self.dependency}: {self.in_image or 'missing'} in {self.image} image, {self.in_native or 'missing'} in Native image"
 
+def find_library_line_number(lib_name: str, file_path: Path):
```
dorschw, line 55:
```suggestion
def find_library_line_number(lib_name: str, file_path: Path) -> int | None:
```

utils/compare_dependency_constraints.py
```diff
@@ -49,8 +50,25 @@ class Discrepancy(NamedTuple):
...
+
+    Parameters:
+    - lib_name: The name of the library to search for.
+    - file_path: The directory containing the pyproject.toml or Pipfile.
+
+    Returns:
+    - The line number containing the library name, or None if the library is not found.
+    """
+    with file_path.open('r') as file:
+        for line_number, line in enumerate(file, start=1):  # Start counting from line 1
```
dorschw, line 67:
```suggestion
    for line_number, line in enumerate(file_path.read_text().splitlines(), start=1):  # Start counting from line 1
```
unindent the rest too 
(p.s. there may be a more-itertools method that does it, but doesn't really matter)

utils/compare_dependency_constraints.py
```diff
@@ -93,8 +110,18 @@ def compare_constraints(images_contained_in_native: list[str]):
...
         )
 
-        for discrepancy in discrepancies:
-            print(str(discrepancy))
+    for discrepancy in discrepancies:
+        file_pyproject_path = DOCKER_FOLDER / f"{discrepancy.image}/pyproject.toml"
+        if file_pyproject_path.exists():
+            file_path = file_pyproject_path
+        else:
+            file_path = DOCKER_FOLDER / f"{discrepancy.image}/Pipfile"
```
dorschw, line 118: We should know the exact file path when creating the discrepancy object, let's add it at that step rather than search for it here

utils/compare_dependency_constraints.py
```diff
@@ -93,8 +110,18 @@ def compare_constraints(images_contained_in_native: list[str]):
...
+        line_number = find_library_line_number(discrepancy.dependency, file_path)
```
dorschw, line 119: what happens when it's `None`?

utils/compare_dependency_constraints.py
```diff
@@ -93,8 +110,18 @@ def compare_constraints(images_contained_in_native: list[str]):
...
+
+        print(
+            f"::error file={file_path},line={line_number},endLine={line_number},title=Native Image Discrepancy::{discrepancy}"
```
dorschw, line 122: Now when we're not in a CLI tool anymore, let's format the error message to something like 

> `requests` is set as `*` in `py3-tools image`, but `{'extras': ['socks'], 'version': '*'}` in the native image. This discrepancy may cause issues when running content.

dorschw (commented):
Good job!

docker/py3-native/pyproject.toml
```diff
@@ -45,7 +45,7 @@ parse-emails = "*"
 slack-sdk = {extras = ["optional"], version = "*"}
 pyjwt = {extras = ["crypto"],version = "*"}
 google-api-python-client = "*"
-exchangelib = ">=5.4.1"
+exchangelib = ">=5.4.0"
```
JudahSchwartz, line 48: changed on purpose?

JudahSchwartz (commented):
looks good, we should add a similiar check that python3-tools and python3-tools-ubi support each other

docker/py3-native/pyproject.toml
shmuel44: yes for testing, I will revert

dorschw:
```suggestion
exchangelib = ">=5.4.1"
```

PR #30231, opened 2024-07-24
update py3-tools-ubi

adi88d:
- updated the base docker to demisto/py3-tools-ubi:1.0.0.105970
- remove the packages: webkit2gtk3-jsc bind-libs.x86_64 polkit polkit-libs


shmuel44 (approved):
Nice

PR #30861, opened 2024-08-06
[Naitve]  `openssl-devel --nobest`

shmuel44:
Because of this issue we added the `--nobest` flag
```
Problem: cannot install the best candidate for the job
nothing provides openssl-libs(x86-64) = 1:3.0.7-27.el9.0.2 needed by openssl-devel-1:3.0.7-27.el9.0.2.x86_64 from rocky-release
```


adi88d (approved):
nice

PR #31404, opened 2024-08-31
Create candidate image version 8.9

shmuel44:
## Related Issues
Related: [link](https://jira-dc.paloaltonetworks.com/browse/CIAC-11663) to the issue

## Description
Create candidate image version 8.9


docker/py3-native/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+## 8.8.0
```
dorschw, line 5: don't we need 8.9 here as well?

shmuel44: Nope we only releasing candidate 8.9

PR #34984, opened 2025-01-07
dockerfile - create a new/update a docker for the dockerfile build

MLainer1:
## Status

## Related Issues
Related: link to the issue

## Description


docker/gitlab-docker-ci/Dockerfile
```diff
@@ -1,3 +1,33 @@
...
+# Install Pyenv and other init steps
+RUN curl -s https://pyenv.run | bash \
+ && eval "$(pyenv init --path)" \
+ && eval "$(pyenv init -)" \
+ && eval "$(pyenv virtualenv-init -)"
+
+# Setup pyenv
+RUN pyenv install -s 2.7.18 && pyenv versions | grep 2.7.18 && \
+ pyenv install -s 3.8.15 && pyenv versions | grep 3.8.15 && \
+ pyenv install -s 3.9.16 && pyenv versions | grep 3.9.16 && \
```
Shellyber, line 26: Why do we need those old versions? I don't think we even support it in our repo anymore

JudahSchwartz: 3.9 i think there might be an image or two that are still stuck on it.
@yaakovpraisler should we remove 3.8 and down?

docker/gitlab-docker-ci/verify.py
```diff
@@ -0,0 +1,18 @@
+import subprocess
+import sys
+
+def verify_python_version(version):
+    """ Verify if provided python version is correctly installed. """
+    try:
+        subprocess.run([f"/root/.pyenv/shims/python{version}", "-c", "import sys; print(sys.version)"], check=True, text=True, stdout=subprocess.PIPE)
```
JudahSchwartz, line 7: i think a better check would be to run pyenv list versions or something to check we can access them through pyenv

docker/gitlab-docker-ci/Dockerfile
Shellyber: Maybe it is worth to remove 3.9 as well (since we're in the process of removing the support for it.

yaakovpraisler: @JudahSchwartz We've discussed it in the KO with @kobymeir and it was decided that we do want them as it was before the migration to gitlab.

yaakovpraisler (commented):
Nice work!
Approved from my side ✅

JudahSchwartz (approved):
great work!

Shellyber (approved):
Nice one!

MLainer1:
This PR needs a force merge due to CVEs found 
image

PR #35329, opened 2025-01-15
Fix update base images

yaakovpraisler:
## Status
Ready

## Related Issues
Fixes: [link to the issue](https://jira-dc.paloaltonetworks.com/browse/CIAC-12547)

## Description
Fix the update-base-images internal/external flows.
Addina the `--python=$(which python)` solves the issue.


yaakovpraisler:
Example [internal](https://github.com/demisto/dockerfiles/actions/runs/12784665506) and [external](https://github.com/demisto/dockerfiles/actions/runs/12784843577) flows

utils/auto_dockerfile_update/update_dockerfiles.py
```diff
@@ -156,7 +156,7 @@ def run_lock(base_path_docker: str, pipfile_or_pyproject_path: str) -> bool:
         if "Pipfile" in pipfile_or_pyproject_path:
             # waits for the process to end.
             completed_process = subprocess.run(
-                ["pipenv", "lock", "--keep-outdated"],
+                ["pipenv", "lock"],
```
JudahSchwartz, line 159: wont this update all the dependencies? Why do we want to do that?

.github/workflows/update-internal-base-images.yml
```diff
@@ -24,8 +24,8 @@ jobs:
             git config --global user.name "auto dockerfiles update"
             echo "==== $(date): Starting pipenv setup... ===="
             python -m pip install --upgrade pip
-            pip install pipenv==2023.3.18
-            pipenv install
+            pip install pipenv==2024.4.0
+            pipenv install --python=$(which python)
```
JudahSchwartz, line 28: why is this flag needed

.github/workflows/update-internal-base-images.yml
```diff
@@ -24,8 +24,8 @@ jobs:
             git config --global user.name "auto dockerfiles update"
             echo "==== $(date): Starting pipenv setup... ===="
             python -m pip install --upgrade pip
-            pip install pipenv==2023.3.18
-            pipenv install
+            pip install pipenv==2024.4.0
```
JudahSchwartz, line 27: do we need to specify a version at all?

utils/auto_dockerfile_update/update_dockerfiles.py
yaakovpraisler: Yes.
Because there is no other solution... those arguments are not supported and I checked this on 4-5 dockers and their build passes ok even if all the dependencies are updated.

.github/workflows/update-internal-base-images.yml
yaakovpraisler: Because otherwise it fails on [this](https://github.com/demisto/dockerfiles/actions/runs/12629465158/job/35187419878)

utils/auto_dockerfile_update/update_dockerfiles.py
JudahSchwartz: there will be an issue for dependencies we specifically have ignored in dependabot config, and major updates will happen without review

.github/workflows/update-internal-base-images.yml
yaakovpraisler: Yes, because we're testing it on a specific version and if tomorrow there will be a new version for pipenv with BCs it will be broken again.

JudahSchwartz (approved):

PR #35877, opened 2025-01-22
Deprecate Trustar

yaakovpraisler:
## Status
Ready


docker/trustar/build.conf
```diff
@@ -1,3 +1,5 @@
 version=20.2.0
+deprecated=true
+deprecated_reason=Image not in use by non-deprecated content item.
```
JudahSchwartz, line 3: you need to update the json file too, the build should fail for it

JudahSchwartz (approved):

PR #38370, opened 2025-07-03
feat: :sparkles: add support for the pyTenable library

BohumilFiala:
## Description
Add support for the pyTenable library.


merit-maita:
@BohumilFiala i see, i will discuss your contribution with my TL and get back to you ASAP.

merit-maita:
Hello @BohumilFiala 
so after looking into your pr, changing the docker image of tenable to use the new pytenable image sounds like a big change. On the other side, it's not common to accept a new docker to be merged before having a content PR, but if you do need this image for developing  your changes I think we can provide you a dev image to use.
 in conlusion, i can't accept this contribution, but i can open a pr on my side with a dev image for you to use.
 please let me know what's your decision.

BohumilFiala:
Hi @merit-maita,

Sorry for the delayed response. If I understand correctly, each newly created Demisto application image has to be linked to a content PR in order to be approved. I'm just trying to understand the process for future PRs.

In the meantime, I contacted our customer and we agreed that, in this case, the private registry in their environment would be the best solution, as we don't plan to share the extension that will be built on the basis of this application image.

I think we can close this PR, as I don't believe the developer image will be useful to us.

Thank you for your assistance.

PR #39706, opened 2025-08-21
Add a new docker image for mjml support.

schroeder247:
MJML is an open-source framework designed to
simplify the process of creating responsive HTML emails. Implemented with mjml-python.

## Status
Ready

## Description
This provides rendering of mjml using mjml-python. A python wrapper of mrml a rust implementation.
The goal is to provide the ability to render mjml in the container instead of using REST APIs keeping potentially sensitive data on-prem. mjml is a framework for responsive email templates.


BEAdi:
Hi @schroeder247, thank you for contributing!
Do you have any contribution planned to https://github.com/demisto/content, that this Docker image will be used in?

schroeder247:
> Hi @schroeder247, thank you for contributing! Do you have any contribution planned to https://github.com/demisto/content, that this Docker image will be used in?
@BEAdi Hi! Yes I will create an integration for this.

BEAdi:
@schroeder247 Which integration are you planning to contribute, and do you intend to join as a partner?
We want to make sure this library will be used, so in the meantime we can provide you with a development Docker image to work with until you're ready to open your content PR.

docker/mjml/Dockerfile
```diff
@@ -0,0 +1,8 @@
+
+FROM demisto/python3:3.12.11.4508456
```
shmuel44, line 2:
```suggestion
FROM demisto/python3:3.12.11.4819260
```

Hi @schroeder247, please use the latest Python3 tag.

docker/mjml/Pipfile
```diff
@@ -0,0 +1,12 @@
...
+verify_ssl = true
+name = "pypi"
+
+[packages]
+mjml-python = "*"
+
+[dev-packages]
+
+[requires]
+python_version = "3.13"
```
shmuel44, line 12:
```suggestion
python_version = "3.12"
```
The Docker image inherits from the Python 3.12 tag, so please use the corresponding Python version in pip.

docker/mjml/Pipfile.lock
shmuel44, line 1: Please lock the pipfile with python3.12, use the following command
```sh
docker run --rm -v $(pwd):/app -w /app demisto/python3:3.12.11.4819260  sh -c "pip install pipenv && pipenv lock --clear --python /usr/local/bin/python3.12"
```

schroeder247:
@BEAdi Yes I would become a partner. I am looking to make an integration that renders mjml. Mjml handles the tricky parts of what is supported in different email clients. The input would be mjml and the output is email client friendly html. XSOAR can be used to automate a lot of things and we need custom styles in certain cases. This eliminates having to write code that's friendly to the nuances of what is supported in each client.

PR #40330, opened 2025-09-15
deprecated unused Docker images

shmuel44:
## Description
deprecated `dockerbuild`, `ami-management`, `py3-tools-ubi`, and `python3-ubi` because we don't use them anymore


BEAdi (approved):

shmuel44:
Hi @BEAdi
This PR needs a force merge, as this image failed to build, but we don't need it to be built, as this is a deprecated image

PR #46298, opened 2026-08-27
fix(sklearn): fail build when NLTK data download fails

shmuel44:
## Related Issues

fixes: XSUP-75716
## Description

To reproduce the bug
```sh
docker run --rm --platform linux/amd64 demisto/sklearn:1.0.0.10866354 \   
  python -c "from nltk.tokenize import sent_tokenize; print(sent_tokenize('Hello there. General Kenobi.'))"
```

The demisto/sklearn:1.0.0.10866354 image shipped with an empty /ml/nltk_data because nltk.download() returns False on failure rather than raising, so a transient CDN failure during the build committed a broken layer and still exited 0. verify.py only imported nltk and never touched the corpora, so the empty data dir passed verification.

Retry each download with backoff and raise on exhaustion, and assert the three resources exist in verify.py.

### DIFF
--- a/docker/sklearn/Dockerfile
+++ b/docker/sklearn/Dockerfile
@@ -17,7 +17,20 @@ RUN apk --update add --no-cache --virtual .build-dependencies python3-dev build-
   && apk del .build-dependencies
 
 RUN mkdir /ml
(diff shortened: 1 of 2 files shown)


BEAdi (commented):
Looks good. Will approve once we prove that this solved the problem to the customer

BEAdi (approved):

PR #46313, opened 2026-08-30
 upgrade requests-ntlm to 1.3.0 to fix MD4 failure under OpenSSL 3.x

shmuel44:
## Related Issues

fixes: XSUP-75734

### Problem

`docker/ntlm`, `docker/py-ews` and `docker/py3ews` pinned `requests-ntlm==1.1.0`,
which pulls the deprecated `ntlm-auth`. It calls `hashlib.new('md4')` while
building the NTLM Type-3 message, and the base image ships OpenSSL 3.x where MD4
sits in the unloaded legacy provider:

    ValueError: unsupported hash type md4

Varonis DSP is fully non-functional (NTLM runs on every command, including
`test-module`); CyberArk CCP pins the same tag. Reproduce:

    docker run --rm demisto/ntlm:1.0.0.9067966 \
      python -c 'from ntlm_auth.compute_hash import _ntowfv1; _ntowfv1("pass")'

The pin cited requests-ntlm #136, a **1.2.0-specific** regression fixed in 1.3.0.
Stale for ~2 years.

### Fix

Upgrade all three to `requests-ntlm==1.3.0`, which uses pyspnego's pure-Python MD4.
Drops the direct `ntlm-auth` dep from both EWS images and removes the three void
dependabot ignore rules that let this rot.

**The `pyspnego>=0.5.2` floor is load-bearing** - 1.3.0 declares only `>=0.4.0`, but
0.5.1 still calls `hashlib.new("md4")`. Without it a resolver could reintroduce the bug.

`py3ews/verify.py` asserted on hashlib MD4 and imported `ntlm_auth`, which would now
fail the build. Replaced with `assert ntowfv1("pass").hex() == "36aa83bd..."`.


docker/ntlm/Pipfile
```diff
@@ -6,7 +6,8 @@ verify_ssl = true
 [dev-packages]
 
 [packages]
-requests-ntlm = "==1.1.0" # lock because of https://github.com/requests/requests-ntlm/issues/136
+pyspnego = ">=0.5.2" # <0.5.2 calls hashlib.new('md4'), fails under OpenSSL 3.x (XSUP-75734)
+requests-ntlm = "==1.3.0"
```
BEAdi:
```suggestion
requests-ntlm = ">=1.3.0"
```

docker/ntlm/verify.py
```diff
@@ -0,0 +1,7 @@
+import requests_ntlm
```
BEAdi, line 2: Should we do something with `requests_ntlm`?

docker/ntlm/verify.py
```diff
@@ -0,0 +1,7 @@
...
+from spnego._ntlm_raw.crypto import ntowfv1
+
+# Make sure NTLM hashing works without OpenSSL's md4 (XSUP-75734):
+assert ntowfv1("pass").hex() == "36aa83bdcab3c9fdaf321ca42a31c3fc"
```
BEAdi, line 9: Is there any other usage of the integration code that can be added here to enhance the verify script?
Same comment for the `py3ews` verify

docker/py-ews/Pipfile
```diff
@@ -22,15 +22,15 @@ idna = "*"
...
 pycparser = "*"
 pygments = "*"
 pykerberos = "*"
+pyspnego = ">=0.5.2" # <0.5.2 calls hashlib.new('md4'), fails under OpenSSL 3.x (XSUP-75734)
 python-dateutil = "*"
 pytz = "*"
 requests = ">=2.32.3"
 requests-kerberos = "*"
-requests-ntlm = "==1.1.0" # lock because of https://github.com/requests/requests-ntlm/issues/136
+requests-ntlm = "==1.3.0"
```
BEAdi:
```suggestion
requests-ntlm = ">=1.3.0"
```

docker/py3ews/Dockerfile
```diff
@@ -8,12 +8,3 @@ RUN apk --update add --no-cache libxslt-dev \
   && apk --update add --no-cache --virtual .build-dependencies python3-dev gcc build-base wget git libffi-dev openssl-dev python3-dev libxml2-dev \
   && pip install --no-cache-dir -r requirements.txt \
   && apk del .build-dependencies
-# Handling the issue described here: https://github.com/ecederstrand/exchangelib/issues/608
```
BEAdi, line 11: Does this show elsewhere in the repo?

docker/py3ews/Pipfile
```diff
@@ -13,8 +13,8 @@ dateparser = "*"
 dnspython = "*"
 enum34 = "*"
 future = ">=1.0.0"
-requests-ntlm = "==1.1.0" # lock because of https://github.com/requests/requests-ntlm/issues/136
-ntlm-auth = ">=1.5.0"
+pyspnego = ">=0.5.2" # <0.5.2 calls hashlib.new('md4'), fails under OpenSSL 3.x (XSUP-75734)
+requests-ntlm = "==1.3.0"
```
BEAdi:
```suggestion
requests-ntlm = ">=1.3.0"
```

PR #46374, opened 2026-09-02
Fix MD4 unsupported hash type in py-ews for NTLM auth

aaron1535:
## Motivation / Background
This PR fixes issue XSUP-76118 where NTLM authentication in EWS v2 fails with `ValueError: unsupported hash type md4`. 

The `py-ews` image relies on OpenSSL 3.x (Debian-based), which moved MD4 to the legacy provider. Since the legacy provider was commented out by default in `/etc/ssl/openssl.cnf`, the `ntlm-auth` library failed instantly when attempting to generate the NT hash for authentication.

## Changes
* **Dockerfile:** Added `sed` commands to uncomment and activate the `legacy_sect` and `default_sect` providers in `/etc/ssl/openssl.cnf`. Added strict `grep` checks to fail the build if the configuration changes in the future base image.
* **verify.py:** Added an assertion to check that `md4` is present in `hashlib.algorithms_available` and explicitly validated the hash output. This ensures CI will automatically block any future base-image updates that break MD4 availability.

## Related Content PR
* **Content PR:** https://github.com/demisto/content/pull/45759

fixes: [XSUP-76118](https://jira-dc.paloaltonetworks.com/browse/XSUP-76118)

### DIFF
--- a/docker/py-ews/Dockerfile
+++ b/docker/py-ews/Dockerfile
@@ -13,6 +13,18 @@ RUN apt-get update && apt-get -y --no-install-recommends upgrade \
 
 RUN sed -i 's/SECLEVEL=2/SECLEVEL=0/g' /etc/ssl/openssl.cnf
 
+# Handling the issue described here: https://github.com/ecederstrand/exchangelib/issues/608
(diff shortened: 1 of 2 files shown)


AradCarmi (approved):
Great!

